PR #3, opened 2020-10-22
instance args

jmchapman:
- [X] (first attempt at) converting instances args to constraints in type sigs
- [x] not expanding instance usage in clause bodies
- [x] type classes
- [ ] instances
- [ ] Dealing with typically derived stuff like Show

PR #4, opened 2020-10-22
Use haskell-src-exts instead of generating strings

UlfNorell:
Also fixes putting LANGUAGE pragmas before the module header.

The code still isn't very nice but at least now we don't have to worry about precedences.

Generates some conflicts for #3.

### DIFF
--- a/Main.hs
+++ b/Main.hs
@@ -52,7 +52,7 @@ pragmaName :: String
 pragmaName = "AGDA2HS"
 
 type Env         = ()
-type ModuleEnv   = ()
+type ModuleEnv   = Builtins
 type ModuleRes   = ()
 type CompiledDef = [(Range, [Hs.Decl ()])]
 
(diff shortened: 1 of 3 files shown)


jmchapman (approved):
On first glance this is great.

A couple of questions:
1. If you wanted to further beautify the haskell source would you then run it through something else?
2. `haskell-src-exts` has this ominous message on hackage:
```
Maintenance
Dan Burton is currently keeping haskell-src-exts on life support. If you are interested in more actively making improvements to this package, please make your interests known.

You might want to try ghc-lib-parser instead.
```
Is `ghc-lib-parser` similar or is the current version of `haskell-src-exts` exactly what we want?

UlfNorell:
Ooh, that's slightly worrying. Merging anyway since it should be a lot easier to switch from `haskell-src-exts` to something else than from the horrible string hack we had before.

I haven't looked at `ghc-lib-parser`.

PR #9, opened 2020-10-25
Travis setup

### LINKED ISSUE #8: Set up github actions to build and run tests
or Travis I guess?

omelkonian:
* closes #8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 dist
 dist-newstyle
 test/build
+test/agda2hs
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+sudo: true
+language: haskell
+cabal: 3.0
+ghc:
+  - "8.6.4"
+
+cache:
+  directories:
+  - "$HOME/.cabal/store"
+
+install:
+  - ghc --version
+  - cabal --version
+  - cabal new-update
+  - travis_wait 60 make install
+
+script:
+  - export PATH=$PATH:$HOME/.cabal/bin
+  - make test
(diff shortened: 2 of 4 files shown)


UlfNorell:
With the changes to `make test` you must now install `agda2hs` globally in order to run the test. This seems like a bad change, since you might not want to do global installs when you are developing something. Is this really necessary for getting it running on travis?

UlfNorell:
Also what do we need to do to see how this actually runs on travis?

jmchapman:
Can we steal something from the Agda travis setup about avoiding global installation of a executable?

omelkonian:
> With the changes to `make test` you must now install `agda2hs` globally in order to run the test. This seems like a bad change, since you might not want to do global installs when you are developing something. Is this really necessary for getting it running on travis?

You are right, we can avoid that by installing on a subfolder prior to testing, and then searching for the executable using `find`.
Let me update this real quick.

omelkonian:
> Also what do we need to do to see how this actually runs on travis?

I have it running in my fork, so you can check all builds/executions on Travis.
Here is the latest one: https://travis-ci.com/github/omelkonian/agda2hs/builds/192734955 (which we will be able to access from a badge in the README).

UlfNorell:
> You are right, we can avoid that by installing on a subfolder prior to testing, and then searching for the executable using find.

Does the `cabal exec` magic I put in there not work on travis?

omelkonian:
> > You are right, we can avoid that by installing on a subfolder prior to testing, and then searching for the executable using find.
> 
> Does the `cabal exec` magic I put in there not work on travis?

Yep, something weird going on with caching.

omelkonian:
New version without the need for a global executable: https://travis-ci.com/github/omelkonian/agda2hs/builds/193418668

UlfNorell (approved):

PR #10, opened 2020-10-27
Extra builtins/primitives: Word/Float/Char 

omelkonian:


test/Test.agda
```diff
@@ -1,76 +1,144 @@
...
+-- chars
+postulate Char : Set
+{-# BUILTIN CHAR Char #-}
+
+primitive
+  primCharToNat : Char → Nat
+  primNatToChar : Nat → Char
+  -- TODO more primitives
+{-# COMPILE AGDA2HS primCharToNat #-}
+{-# COMPILE AGDA2HS primNatToChar #-}
```
UlfNorell, line 108: This isn't really how you want this to work. Anything which doesn't let you simply import `Agda.Builtin.*` is a no-go in my opinion.

omelkonian: Hmm, how would you go about this though? If there is no user PRAGMA to instruct `agda2hs` to compile a primitive, would we internally keep a (fixed) set of primitives and replace all occurrences in terms?

omelkonian: Or something like an RTS.hs, which is imported by default and includes the primitive definitions in Haskell?

UlfNorell: I think we really want to avoid any `RTS.hs` things here. The idea is that you are using Agda to write (and prove things about) a Haskell program, so we'd like the end result to be a Haskell program that you might have written directly in Haskell. Having a special RTS module kind of defeats that.

Given this premise I think this can be handled as a library on the Agda side. The question isn't "How should we translate `primNatToChar` (say)?". The question is "How do we represent `fromEnum :: Char -> Int` in Agda?". 

It's perfectly fine if there are lots of Agda primitives/constructions that we don't translate.

test/Test.agda
```diff
@@ -1,76 +1,144 @@
...
+-- comment
+-- another comment
+bla :: Int -> Int
+bla n = n * 4
+
+{- multi
+   line
+   comment
+-}
 #-}
```
UlfNorell, line 38: There was a reason to put this between two `COMPILE` pragmas: to make sure that the sorting of `FOREIGN` blocks and `COMPILE`s is working.

omelkonian: Oops, totally missed that, I will revert this.

test/Test.agda
omelkonian: How about we postulate a marker that interprets a given HS string to any type and then the backend emits the given string immediately?
```agda
-- ** Extra builtins

postulate
  $ : ∀ {A : Set} → String → A

-- floats
ex_float : Float
ex_float = 0.0
{-# COMPILE AGDA2HS ex_float #-}

sumF : List Float → Float
sumF []       = 0.0
sumF (x ∷ xs) = $ {_ → _ → _} "(+) :: Double -> Double -> Double" x (sumF xs)
{-# COMPILE AGDA2HS sumF #-}

-- chars
ex_char : Char
ex_char = 'x'
{-# COMPILE AGDA2HS ex_char #-}

ex_toℕ : Nat
ex_toℕ = $ {_ → _} "(fromIntegral . fromEnum :: Char -> Integer)" 'd'
{-# COMPILE AGDA2HS ex_toℕ #-}

baseNat : Nat
baseNat = $ {Char → _} "(fromIntegral . fromEnum :: Char -> Integer)" ($ "toEnum 0")
{-# COMPILE AGDA2HS baseNat #-}
```

test/Test.agda
```diff
@@ -1,40 +1,48 @@
...
+
 data Exp (v : Set) : Set where
   Plus : Exp v → Exp v → Exp v
   Int : Nat → Exp v
   Var : v → Exp v
-
 {-# COMPILE AGDA2HS Exp #-}
 
 eval : (a → Nat) → Exp a → Nat
 eval env (Plus a b) = eval env a + eval env b
```
omelkonian, line 36: The translation of `+` troubles me a bit, since it is only a coincidence that our Agda implementation of addition happens to be infix and called exactly the same as Haskell's addition on Nums. 
@UlfNorell somewhat related to the discussion on primitive functions

test/Test.agda
UlfNorell: This looks terrible. The reason to have the code in Agda is that you want to prove something about it. Having a postulate defeats that completely.

What is that actual problem that you want to solve here? I'm not convinced we need to go any further than #12.

omelkonian: Yeah, you're right. Let's archive this PR.

PR #12, opened 2020-10-28
Extra builtins types: Word/Float/Char

omelkonian:
This introduces new built-in types for `Data.Word.Word64`, `Double` and `Char` without any fancy business for adding primitives (one can of course just postulate a function with the same name as the Haskell one).

### DIFF
--- a/Main.hs
+++ b/Main.hs
@@ -15,6 +15,7 @@ import System.Directory
 
 import qualified Language.Haskell.Exts.SrcLoc     as Hs
 import qualified Language.Haskell.Exts.Syntax     as Hs
+import qualified Language.Haskell.Exts.Build      as Hs
 import qualified Language.Haskell.Exts.Pretty     as Hs
 import qualified Language.Haskell.Exts.Parser     as Hs
 import qualified Language.Haskell.Exts.ExactPrint as Hs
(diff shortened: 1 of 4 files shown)


Main.hs
```diff
@@ -111,39 +112,51 @@ isOp _                         = False
...
-compilePrim Nat  = Hs.UnQual () (hsName "Integer")
-compilePrim List = Hs.Special () (Hs.ListCon ())
-compilePrim Unit = Hs.Special () (Hs.UnitCon ())
-compilePrim Cons = Hs.Special () (Hs.Cons ())
-compilePrim Nil  = Hs.Special () (Hs.ListCon ())
+compilePrim = \case
+  Unit   -> special Hs.UnitCon
+  Nat    -> unqual "Integer"
+  Float  -> unqual "Double"
+  Word64 -> qual "Data.Word" "Word64"
```
UlfNorell, line 144: Having this qualified isn't very nice in the generated code. You anyway have to import `Data.Word` even to get the qualified version, so I see no strong reason to not simply use `Word64`.

UlfNorell (changes requested):

Main.hs
omelkonian: It would be nice if we automatically inserted the necessary imports for built-in types, what do you think?

UlfNorell: Yeah, why not. We are parsing the user imports, so it should be easy enough to check if it's already there. You probably also want to check that you don't have `Word64` explicitly imported from another module. For checking whether or not it's used I guess the easiest solution is to traverse the generated Haskell code.

UlfNorell (approved):
Nice

PR #13, opened 2020-10-28
Compile if_then_else_ expressions 

omelkonian:


Main.hs
```diff
@@ -229,18 +246,28 @@ compileTerm :: Builtins -> Term -> TCM (Hs.Exp ())
 compileTerm builtins v =
   case unSpine v of
     Var x es   -> (`app` es) . Hs.Var () . Hs.UnQual () . hsName =<< showTCM (Var x [])
-    Def f es   -> (`app` es) . Hs.Var () =<< hsQName builtins f
+    Def f es  
+      | show (qnameName f) == "if_then_else_"
+      -> do [b, left, right] <- compileArgs es
```
UlfNorell, line 251: Here we should maybe fail more gracefully (or eta-expand!) on partially applied `if`, and not fail at all on over-application. For instance
```haskell
(if b then id else suc) 0
```

test/Test.agda
```diff
@@ -1,40 +1,55 @@
...
+open import Agda.Builtin.Bool
+open import Agda.Builtin.Float
+open import Agda.Builtin.Word
+open import Agda.Builtin.Char
 open import Agda.Builtin.Equality
 
 variable
   a b : Set
 
+-- TODO: move to Prelude.hs?
```
UlfNorell, line 14: The test suite is set up to handle multiple files so I think it makes sense to split up this file into smaller pieces. (Just remember to make all modules reachable from `AllTests.agda`).

UlfNorell (changes requested):

Main.hs
```diff
@@ -110,12 +110,25 @@ isOp (Hs.UnQual _ Hs.Symbol{}) = True
 isOp (Hs.Special _ Hs.Cons{})  = True
 isOp _                         = False
 
+freshString :: String -> TCM String
+freshString s = do
+  ctx <- getContextNames
+  -- prettyShow <$> freshName_ s
+  -- ^ this returns 'x' even when context contains 'x'??
```
omelkonian, line 117: Am I using `freshName_` improperly here?

UlfNorell: Try `prettyTCM` instead of `prettyShow`.

Main.hs
```diff
@@ -241,7 +258,11 @@ compileTerm :: Builtins -> Term -> TCM (Hs.Exp ())
 compileTerm builtins v =
   case unSpine v of
     Var x es   -> (`app` es) . Hs.Var () . Hs.UnQual () . hsName =<< showTCM (Var x [])
-    Def f es   -> (`app` es) . Hs.Var () =<< hsQName builtins f
+    Def f es
+      | show (qnameName f) == "if_then_else_"
+      -> ite es
+      | otherwise
+      -> (`app` es) . Hs.Var () =<< hsQName builtins f
```
UlfNorell, line 265:
```suggestion
      | show (qnameName f) == "if_then_else_" -> ite es
      | otherwise -> (`app` es) . Hs.Var () =<< hsQName builtins f
```

Main.hs
```diff
@@ -251,11 +272,23 @@ compileTerm builtins v =
...
+    app hd es = eApp <$> pure hd <*> compileArgs es
+
+    ite :: Elims -> TCM (Hs.Exp ())
+    ite es = compileArgs es >>= \case
+      -- fully applied
+      (b : t : f : es') -> return $ Hs.If () b t f `eApp` es'
+      -- partially applied -> eta-expand
+      es' -> do
+        xs <- fmap Hs.name . drop (length es') <$> mapM freshString ["b", "t", "f"]
+        let [b, t, f] = es' ++ (Hs.var <$> xs)
```
UlfNorell, line 290:
```suggestion
        let [b, t, f] = es' ++ map Hs.var xs
```

Prelude.agda
```diff
@@ -0,0 +1,17 @@
...
+open import Agda.Builtin.Char public
+open import Agda.Builtin.Equality public
+
+variable
+  a b c d e f g h i j k l m n o p q r s t u v w x y z : Set
+
+infix 0 if_then_else_
+if_then_else_ : Bool → a → a → a
+if true  then t else f = t
+if false then t else f = f
```
UlfNorell, line 17: Committed by mistake?

UlfNorell (changes requested):

Main.hs
omelkonian: Aha, it works!

Prelude.agda
omelkonian: We need it because it's not defined in Agda.Builtin.Bool, no?

UlfNorell: I meant this copy of `Prelude.agda` in the root. The one in `test/` is fine.

UlfNorell (approved):

UlfNorell:
👍

PR #16, opened 2020-10-31
Automatic builtin imports + checking them

omelkonian:
* Moves Haskell utilities to separate `HsUtils.hs` file

  * Also adds quasiquoter for `hs-src-exts` things (spliced in expressions/patterns)

### DIFF
--- a/Main.hs
+++ b/Main.hs
@@ -58,8 +58,6 @@ import HsUtils
 data Options = Options { optOutDir     :: FilePath,
                          optExtensions :: [Hs.Extension] }
 
-data Options = Options { outDir :: FilePath }
-
 defaultOptions :: Options
 defaultOptions = Options{ optOutDir = ".", optExtensions = [] }
 
@@ -384,7 +382,7 @@ imports modules = concat [imps | (_, (Hs.Module _ _ _ imps _, _)) <- modules]
 
 addImports :: [Hs.ImportDecl Hs.SrcSpanInfo] -> [CompiledDef] -> TCM [Hs.ImportDecl ()]
 addImports is defs = do
-  return [importWord64 | not (null defs || any isImportWord64 is)] 
+  return [importWord64 | usesWord64 defs && not (any isImportWord64 is)]
   where
     importWord64 :: Hs.ImportDecl ()
     importWord64 = Hs.ImportDecl ()
--- a/Main.hs
+++ b/Main.hs
@@ -28,7 +28,7 @@ import Agda.Compiler.Common
 import Agda.Interaction.BasicOps
 import Agda.TypeChecking.Monad
 import Agda.TypeChecking.Pretty hiding (pretty)
-import Agda.Syntax.Common
+import Agda.Syntax.Common hiding (Ranged)
 import qualified Agda.Syntax.Concrete.Name as C
 import Agda.Syntax.Literal
 import Agda.Syntax.Internal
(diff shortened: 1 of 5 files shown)


UlfNorell:
Pulling in template haskell quasi quoting just for one import statement seems a bit excessive. Especially since it compllicates building with different ghc versions. Do you see it being more useful in other cases?

omelkonian:
> Pulling in template haskell quasi quoting just for one import statement seems a bit excessive. Especially since it compllicates building with different ghc versions. Do you see it being more useful in other cases?

You are absolutely right this is a bit of an overkill (Travis is already failing :P) . Since the quasi-quoter only supports constant terms, we will probably make little use of it. 

However, if it could support meta-vars/anti-quotation it would become quite useful in all our transformations, so I will shelve this for now, until I can offer anti-quotation as well (probably it's not possible/feasible, otherwise `hs-src-exts` would provide a quasi-quoter already?).

Main.hs
```diff
@@ -363,29 +315,69 @@ renderBlocks = unlines . map unlines . sortRanges . filter (not . null . snd)
...
-        sub :: Int -> Hs.SrcSpan -> Hs.SrcSpan
-        sub l (Hs.SrcSpan f l0 c0 l1 c1) = Hs.SrcSpan f (l0 - l) c0 (l1 - l) c1
+imports :: [Ranged Code] -> [Hs.ImportDecl Hs.SrcSpanInfo]
+imports modules = concat [imps | (_, (Hs.Module _ _ _ imps _, _)) <- modules]
 
-pp :: Hs.Pretty a => a -> String
-pp = Hs.prettyPrintWithMode Hs.defaultMode{ Hs.spacing = False }
+addImports :: [Hs.ImportDecl Hs.SrcSpanInfo] -> [CompiledDef] -> TCM [Hs.ImportDecl ()]
+addImports is defs = do
+  return [importWord64 | not (null defs || any isImportWord64 is)]
```
omelkonian, line 335: This will add `import Data.Word (Word64)` to every module with a non-empty set of (Haskell) declarations, but we probably want to do that only when there is relevant code in that module. This seems quite hard to check though (you need a hundred LOC to traverse all possible `Hs.`constructors and check types referred to there), but maybe a SYB-generic function can do that easily enough.

UlfNorell: Another option is to remember when you generate the `Word64` type.

I think if you can't detect when to import it, it's better to not do anything than to always import. It's not that hard to write the Haskell import yourself and it would be really annoying to have an unnecessary `Data.Word` import everywhere.

omelkonian: OK, let's try to do this on the Agda side (i.e. remembering when we generate `Word64`) and if it's not so easy, abandon automatic importing altogether. Any quick suggestion on how to manage custom state in `TCM`, do I really need to wrap it in `WriterT` or is there a convenient way to do this from TCM itself?

UlfNorell: Yeah, I think `WriterT` is your best bet.

omelkonian: Phew, there was a trivial way to do this using SYB's [listify](https://hackage.haskell.org/package/syb-0.7.1/docs/Data-Generics-Schemes.html#v:listify).

UlfNorell (approved):
Nice! Go ahead and merge.

UlfNorell:
Caching worked perfectly here. `syb` was installed and nothing else was touched 🎉

PR #18, opened 2020-11-02
Fix #14

### LINKED ISSUE #14: Name conflict when translating sections
When mixfix operators have unordered sections filled in, Agda generates a binding of a variable named `section`, which may clash with user-provided identifiers of the same name.

As an example:
```agda
inc : List Nat -> List Nat
inc = map (_+ 1)
{-# COMPILE AGDA2HS inc #-}
  -- ~> inc = map (\ section -> section + 1)

incBy : Nat -> List Nat -> List Nat
incBy section = map (_+ section)
{-# COMPILE AGDA2HS incBy #-}
  -- ~> incBy section = map (\ section -> section₁ + section)
```

UlfNorell:
The scope needs to be updated when going under binders to force the printer to not rename user-written variables.

Fixes #14

### DIFF
--- a/test/Issue14.hs
+++ b/test/Issue14.hs
@@ -1,0 +1,11 @@
+module Issue14 where
+
+constid :: a -> b -> b
+constid x = \ x -> x
+
+sectionTest₁ :: Integer -> Integer -> Integer
+sectionTest₁ n = \ section -> section + n
+
+sectionTest₂ :: Integer -> Integer -> Integer
+sectionTest₂ section = \ section -> section + section₁
+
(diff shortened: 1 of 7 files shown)


test/Issue14.hs
```diff
@@ -0,0 +1,11 @@
...
+
+constid :: a -> b -> b
+constid x = \ x -> x
+
+sectionTest₁ :: Integer -> Integer -> Integer
+sectionTest₁ n = \ section -> section + n
+
+sectionTest₂ :: Integer -> Integer -> Integer
+sectionTest₂ section = \ section -> section + section₁
+
```
omelkonian, line 11: This seems to have been committed by mistake, right? (it's also an incorrect compilation `!= test/golden/Issue14.hs`

UlfNorell: Oh, well spotted. Must have been left over from one of the test runs.

PR #19, opened 2020-11-02
Switch to github actions

UlfNorell:
cc #8

- [x] PR to see if it works... First run worked.
- [x] Gratuitous force push to see if the caches works. Seems to have worked. All three tests ran in < 2min.
- [x] ~~Getting warnings about Cabal-3.2, trying with 3.4 instead which is listed as pre-installed. That (unsurprisingly) invalidated the caches. Still doesn't work after successful run and force push!~~
- [x] Trying with bumped versions of actions and `Cabal: latest`. That worked, but latest resolved to `3.2`, so the warning is back.
- [x] ~~Trying again with `3.4`. Still doesn't work, but now I'm really confused because there isn't a `cabal 3.4` on Hackage?! They must have a 3.4 release candidate preinstalled.~~
- [x] Trying with `3.4` and `restore-keys`. Success!

I guess the problem with 3.4 didn't go away though, it still gets a different fingerprint every time. But with `restore-keys` the only effect of that is that we upload a new cache for every build.


UlfNorell:
Here's what's causing the cache miss: `cabal-3.4` includes the hackage index state in the freeze file:
```
index-state: hackage.haskell.org 2020-11-03T05:53:25Z
```
I suppose that makes some kind of sense. It shouldn't change all the time so this is probably what you want.

omelkonian (approved):
LGTM

omelkonian:
> Here's what's causing the cache miss: `cabal-3.4` includes the hackage index state in the freeze file:
> 
> ```
> index-state: hackage.haskell.org 2020-11-03T05:53:25Z
> ```
> 
> I suppose that makes some kind of sense. It shouldn't change all the time so this is probably what you want.

Maybe we could have a `cabal.project` file that sets a particular `index-state`, so we have more control over cache misses?
e.g. in `agda2hs/cabal.project`
```yaml
repository head.hackage
  url: http://head.hackage.haskell.org/
  secure: True
  root-keys: 07c59cb65787dedfaef5bd5f987ceb5f7e5ebf88b904bbd4c5cbdeb2ff71b740
             2e8555dde16ebd8df076f1a8ef13b8f14c66bad8eafefd7d9e37d0ed711821fb
             8f79fd2389ab2967354407ec852cbe73f2e8635793ac446d09461ffb99527f6e
  key-threshold: 3

index-state: 2020-11-03T05:53:25Z
```

UlfNorell:
Not sure it's worth it to be honest. That would need to be updated periodically and the only effect a cache miss has is that we upload a new cache, which takes no more than a few seconds.

UlfNorell:
The job on `master` didn't find a cache! I guess the caches must be local to each branch? I really hope new PRs can use the `master` cache.

UlfNorell:
Yes it seems so. From https://github.com/actions/cache:

> The cache is scoped to the key and branch. The default branch cache is available to other branches.

omelkonian:
> The job on `master` didn't find a cache! I guess the caches must be local to each branch? I really hope new PRs can use the `master` cache.

It should be similar to Travis, which keeps a cache for each branch/PR, but can always fallback to the 'default' cache (the one in `master`): https://docs.travis-ci.com/user/caching#fetching-and-storing-caches.

Also, from the Github CI docs: https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

PR #20, opened 2020-11-03
Refactor builtins and add support for tuples

UlfNorell:
- Start on a Haskell.* library meant to mimic selected parts of the Haskell base library
- Builtins based on fully qualified names instead of the Agda builtin mappings
- Generic tuple type in Haskell.Prelude mapped to Haskell tuples


Main.hs
```diff
@@ -113,100 +113,169 @@ hsQName builtins f
...
+  xs <- makeList' "Haskell.Prelude.Tuple.[]" "Haskell.Prelude.Tuple._∷_" err v
+  ts <- mapM compileTerm xs
+  return $ Hs.Tuple () Hs.Boxed ts
+
+tuplePat :: ConHead -> ConPatternInfo -> [NamedArg DeBruijnPattern] -> TCM (Hs.Pat ())
+tuplePat cons i ps = do
+  let p = ConP cons i ps
+  err <- sep [ text "Tuple pattern"
+             , nest 2 $ prettyTCM p
+             , text "does not have a known size." ]
```
omelkonian, line 225: Can we add an error-expectation test for this, or is it simply `__IMPOSSIBLE__`? (and similarly for `tupleTerm`)

Main.hs
```diff
@@ -113,100 +113,169 @@ hsQName builtins f
...
+  xs <- makeListP' "Haskell.Prelude.Tuple.[]" "Haskell.Prelude.Tuple._∷_" err p
+  qs <- mapM compilePat xs
+  return $ Hs.PTuple () Hs.Boxed qs
 
 -- Compiling things -------------------------------------------------------
 
-compile :: Options -> Builtins -> IsMain -> Definition -> TCM CompiledDef
-compile _ builtins _ def = getUniqueCompilerPragma pragmaName (defName def) >>= \ case
-  Just _  -> compile' builtins def
+compile :: Options -> () -> IsMain -> Definition -> TCM CompiledDef
```
omelkonian, line 232: For consistency and resilience to future additions, I would put `ModuleEnv` here (instead of `()`).

Main.hs
```diff
@@ -113,100 +113,169 @@ hsQName builtins f
...
+    "Agda.Builtin.Float.Float"     -> unqual "Double"
+    "Agda.Builtin.Bool.Bool.false" -> unqual "False"
+    "Agda.Builtin.Bool.Bool.true"  -> unqual "True"
+    "Agda.Builtin.List.List"       -> special Hs.ListCon
+    "Agda.Builtin.List.List._∷_"   -> special Hs.Cons
+    "Agda.Builtin.List.List.[]"    -> special Hs.ListCon
+    "Agda.Builtin.Unit.⊤"          -> special Hs.UnitCon
+    "Agda.Builtin.Unit.tt"         -> special Hs.UnitCon
+    "Haskell.Prelude.Tuple.[]"     -> special Hs.UnitCon
+    _ -> Nothing
```
omelkonian, line 184: Fantastic, that's much cleaner to work with and no need for the dummy intermediate `Prim` datatype!

Main.hs
UlfNorell: It's not impossible. We haven't set up any negative tests yet, but this is definitely a candidate. Something like
```agda
bla : a -> Tuple as -> Tuple (a :: as)
bla x xs = x :: xs
```
Might be tricky to not hit the error in the type first though.

omelkonian (approved):
LGTM

Main.hs
omelkonian: Note: an artificial example for getting to this error:
```agda
bla : Tuple [] → Tuple (Nat ∷ [])
bla xs = 1 ∷ xs
{-# COMPILE AGDA2HS bla #-}

```

PR #21, opened 2020-11-03
Compile sections

UlfNorell:
~~On top of #20~~

### DIFF
--- a/Main.hs
+++ b/Main.hs
@@ -329,13 +329,9 @@ compileType t = do
     t -> genericDocError =<< text "Bad Haskell type:" <?> prettyTCM t
 
 compileTerm :: Term -> TCM (Hs.Exp ())
-compileTerm v =
+compileTerm v = do
   case unSpine v of
-    Var x es   -> do
-      cxt <- getContext
-      reportSDoc "agda2hs.compile" 10 $ text "context:" <+> fsep
-        [ prettyTCM (Var i []) | i <- reverse [0..length cxt - 1] ]
-      (`app` es) . Hs.Var () . Hs.UnQual () . hsName =<< showTCM (Var x [])
+    Var x es   -> (`app` es) . Hs.Var () . Hs.UnQual () . hsName =<< showTCM (Var x [])
     Def f es
       | Just semantics <- isSpecialTerm f -> semantics f es
       | otherwise -> (`app` es) . Hs.Var () =<< hsQName f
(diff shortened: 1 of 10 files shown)


test/golden/Test.hs
```diff
@@ -54,7 +54,7 @@ map f [] = []
 map f (x : xs) = f x : map f xs
 
 mapTest :: [Integer] -> [Integer]
-mapTest = map (id . (+) 5)
+mapTest = map (id . (5 +))
```
omelkonian, line 57: Given that you just added support for sections, the user could have written `map (id ∘ (5 +_))` in the Agda to get this compilation result. Why impose this behaviour, making it impossible to actually get `(+) 5`?

Main.hs
```diff
@@ -329,13 +329,9 @@ compileType t = do
     t -> genericDocError =<< text "Bad Haskell type:" <?> prettyTCM t
 
 compileTerm :: Term -> TCM (Hs.Exp ())
-compileTerm v =
+compileTerm v = do
```
omelkonian, line 332: This `do` is unnecessary, probably a leftover from debugging?

Main.hs
```diff
@@ -350,10 +346,14 @@ compileTerm v =
...
       -- System-inserted lambda, no need to preserve the name.
-      -- TODO: most likely a section, so should really reconstruct that.
       underAbstraction_ b $ \ body -> do
         x <- showTCM (Var 0 [])
-        hsLambda x <$> compileTerm body
+        let hsx = Hs.Var () $ Hs.UnQual () (hsName x)
+        body <- compileTerm body
+        return $ case body of
+          Hs.InfixApp _ a op b
+            | a == hsx -> Hs.RightSection () op b -- System-inserted visible lambdas can only come from sections
```
omelkonian, line 355: `x` may not be system-inserted from section, so this again imposes sections to the user, e.g.
```agda
test₁ : Nat → Nat
test₁ = λ x → 5 + x
```
compiles to:
```haskell
test₁ :: Integer -> Integer
test₁ = (5 +)
```

test/golden/Test.hs
UlfNorell: The problem is that we cannot distinguish `_+_ 5` and `5 +_`, so I opted for always using the section.

Main.hs
UlfNorell: That's a left section so doesn't happen here. The problem is that Agda eta-contracts the lambda before it gets to us. Right-sections can only come from actual user-written sections.

test/golden/Test.hs
omelkonian: OK, this seems reasonable.

omelkonian (approved):
LGTM

PR #22, opened 2020-11-03
Automatically load first command in repl

omelkonian:


UlfNorell (changes requested):
I also tried (and failed) to get this working. Unfortunately this solution breaks tab completion in the repl :(

omelkonian:
> I also tried (and failed) to get this working. Unfortunately this solution breaks tab completion in the repl :(

That's a bummer, closing this.

PR #23, opened 2020-11-08
Compile where clauses

omelkonian:
* also compiling postulates to undefined

### DIFF
--- a/Main.hs
+++ b/Main.hs
@@ -4,7 +4,7 @@ import Control.Arrow ((>>>))
 import Control.Monad
 import Control.Monad.Fail (MonadFail)
 import Control.Monad.IO.Class
-import Data.Generics (mkT, everywhere)
+import Data.Generics (mkT, everywhere, listify, extT, everything, mkQ)
 import Data.Function
 import Data.List
 import Data.Maybe
@@ -34,7 +34,6 @@ import Agda.Syntax.Common hiding (Ranged)
 import qualified Agda.Syntax.Concrete.Name as C
 import Agda.Syntax.Literal
 import Agda.Syntax.Internal
-import Agda.Syntax.Internal.Pattern (patternsToElims)
 import Agda.Syntax.Position
 import Agda.Syntax.Translation.ConcreteToAbstract
 import Agda.Syntax.Translation.AbstractToConcrete
@@ -148,6 +147,19 @@ underAbstr a b ret = underAbstraction' KeepNames a b $ \ body ->
 
 underAbstr_ :: Subst t a => Abs a -> (a -> TCM b) -> TCM b
 underAbstr_ = underAbstr __DUMMY_DOM__
+
+applyNoBodies :: Definition -> [Arg Term] -> Definition
+applyNoBodies d args = revert $ d `apply` args
+  where
+    bodies0 :: [(QName, [(Int, Maybe Term)])]
+    bodies0 = everything (++) ([] `mkQ` go) d
+      where go (Defn {defName = q, theDef = Function {funClauses = cls}}) =
+              [(q, map (fmap clauseBody) (zip [0..] cls))]
+
+    revert :: Definition -> Definition
+    revert d@(Defn {defName = q, theDef = f@(Function {funClauses = cls})})
+      = d {theDef = f {funClauses = map (\(i, c) -> c {clauseBody = get i}) (zip [0..] cls)}}
+      where get i = fromMaybe __IMPOSSIBLE__ $ join $ lookup i <$> lookup q bodies0
 
 -- Builtins ---------------------------------------------------------------
 
@@ -290,31 +302,26 @@ compilePostulate def = do
 
 type LocalDecls = [(QName, Definition)]
 
-groupDecls :: LocalDecls -> [(Definition, LocalDecls)]
-groupDecls [] = []
-groupDecls ((q, d) : defs) =
-  let getLvl (QName (MName xs) _) = length xs
-      (children, defs') = span ((> getLvl q) . getLvl . fst) defs
-  in (d, children) : groupDecls defs'
-
 compileFun :: Definition -> TCM [Hs.Decl ()]
 compileFun d = do
-  let findLocals = fromMaybe [] . fmap snd . find ((defName d ==) . defName . fst)
-  locals <- findLocals . groupDecls . sortDefs <$> curDefs
-  compileFun' locals d
+  locals <- takeWhile (isAnonymousModuleName . qnameModule . fst)
+          . dropWhile ((<= defName d) . fst)
+          . sortDefs <$> curDefs
+  compileFun' d locals
 
-compileFun' :: LocalDecls -> Definition -> TCM [Hs.Decl ()]
-compileFun' locals def@(Defn {..}) = do
+compileFun' :: Definition -> LocalDecls -> TCM [Hs.Decl ()]
+compileFun' (Defn {..}) locals = do
   let n = qnameName defName
       x = hsName $ prettyShow n
+      go = foldM $ \(ds, ms) -> compileClause ds defName x >=> return . fmap (ms `snoc`)
   setCurrentRange (nameBindingSite n) $ do
     ty <- compileType (unEl defType)
-    cs <- mapM (compileClause locals defName x) funClauses
+    cs <- snd <$> go (locals, []) funClauses
     return [Hs.TypeSig () [x] ty, Hs.FunBind () cs]
   where
     Function{..} = theDef
 
-compileClause :: LocalDecls -> QName -> Hs.Name () -> Clause -> TCM (Hs.Match ())
+compileClause :: LocalDecls -> QName -> Hs.Name () -> Clause -> TCM (LocalDecls, Hs.Match ())
 compileClause locals qn x c@Clause{clauseTel = tel, namedClausePats = ps', clauseBody = body} =
   addContext (KeepNames tel) $ localScope $ do
     -- Compile patterns, only bind user-written variables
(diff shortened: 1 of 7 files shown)


UlfNorell (changes requested):
If you try to add an argument to `ex` and use it in one of the `where` functions it fails. I don't think it's worth doing this if it only works for definitions without left-hand sides. 

Also, do you have a use case in mind for compiling postulates to undefined? I can't think of situation where you would actually want this...

omelkonian:
> If you try to add an argument to `ex` and use it in one of the `where` functions it fails. I don't think it's worth doing this if it only works for definitions without left-hand sides.

I see, the problem is that the definitions inside `where` are 'lifted' and thus require the arguments to `ex` as input again.
Would it be possible to (partially) apply all of these definition and then compile?
Or is this approach doomed to fail from your experience?

> Also, do you have a use case in mind for compiling postulates to undefined? I can't think of situation where you would actually want this...

Re. postulates, you will never want it in the end result, but it seems useful while prototyping:  think of `postulate a : A` as a shorthand for:
```haskell
a :: A
a = undefined
```
(another way to do it is add a postulate in `Prelude.undefined` and use a full declaration)

UlfNorell:
> Would it be possible to (partially) apply all of these definitions and then compile? Or is this approach doomed to fail from your experience?

This should be possible. There's an `Apply` instance for `Definition` so hopefully you can grab the lhs variables of the `clauseTel` and simply `apply` them before compiling the where-functions.

UlfNorell:
> another way to do it is add a postulate in Prelude.undefined and use a full declaration

Hm.. since the objective is to prove things about the code, I'd rather not add `undefined` to the Prelude. Maybe it could be `error "postulate: a"` rather than `undefined` though.

Main.hs
```diff
@@ -273,30 +276,66 @@ compileConstructorArgs (ExtendTel a tel)
...
-    ps   <- compilePats ps
-    body <- maybe (return $ Hs.Var () $ Hs.UnQual () (hsName "undefined")) compileTerm body
-    let rhs = Hs.UnGuardedRhs () body
+    ps <- compilePats ps'
+
+    -- Compile rhs and its @where@ clauses, making sure that:
+    --   * inner modules get instantiated
+    --   * references to inner modules get un-qualifiyed (and instantiated)
+    let elims = patternsToElims ps'
+        children = groupDecls $ map (fmap (`applyE` elims)) locals
```
UlfNorell, line 326: This is not correct. Local functions are parameterised over the variables bound in the lhs, not over the arguments to the parent function. You need `teleArgs tel` (and `apply` instead of `applyE`).

Main.hs
```diff
@@ -273,30 +276,66 @@ compileConstructorArgs (ExtendTel a tel)
...
+groupDecls ((q, d) : defs) =
+  let getLvl (QName (MName xs) _) = length xs
+      (children, defs') = span ((> getLvl q) . getLvl . fst) defs
+  in (d, children) : groupDecls defs'
+
+compileFun :: Definition -> TCM [Hs.Decl ()]
+compileFun d = do
+  let findLocals = fromMaybe [] . fmap snd . find ((defName d ==) . defName . fst)
+  locals <- findLocals . groupDecls . sortDefs <$> curDefs
+  compileFun' locals d
```
UlfNorell, line 302: This does not work for functions with multiple clauses. Or rather, it might, but you need to do a reachability analysis in each clause to make sure you only get the local functions for each individual clause.

test/Where.agda
```diff
@@ -0,0 +1,51 @@
...
+-- nested where with outer arguments
+ex4 : Bool -> Nat
+ex4 b = mult 42
+  where
+    mult : Nat -> Nat
+    mult n = bump 5 (bool2nat b)
+      where
+        bump : Nat -> Nat -> Nat
+        bump x y = x * y + n
+{-# COMPILE AGDA2HS ex4 #-}
```
UlfNorell, line 51: Here's a test case:
```agda
-- with pattern matching and multiple clauses
ex5 : List Nat → Nat
ex5 [] = zro
  where
    zro = 0
ex5 (n ∷ ns) = mult num + 1
  where
    num = 42 + ex5 ns

    mult : Nat → Nat
    mult = _* n
{-# COMPILE AGDA2HS ex5 #-}
```

UlfNorell (changes requested):
Progress! Still doesn't quite work for functions with multiple clauses.

test/Where.agda
omelkonian: Almost there.. I have now come across a rather peculiar behaviour in this compilation:
```agda
ex7 : Nat → Nat
ex7 n₀ = go₁ n₀
  where
    go₁ : Nat → Nat
    go₁ n₁ = go₂ (n₀ ⊗ n₁)
      where
        go₂ : Nat → Nat
        go₂ n₂ = n₀ ⊗ n₁ ⊗ n₂
```
translates to
```hs
ex7 :: Integer -> Integer
ex7 n₀ = go₁ n₀
  where go₁ :: Integer -> Integer
        go₁ n₁ = go₂ (n₀ ⊗ n₁)
          where go₂ :: Integer -> Integer
                go₂ n₂ = n₁ ⊗ n₁ ⊗ n₂
```

After some debugging, the weird DeBruijn index changes when handling clause `go₁ n₁ = ...`, which has to partially apply the inner definition of go₂. While doing that, the indices for the expression `n₀ ⊗ n₁ ⊗ n₂` ~ `@2 @1 @0` change to `@1 @1 @0`. Any guess on what I might be doing wrong here? Here is the relevant part of the log:
```bash
****** Clause ******
(n₁ : Agda.Builtin.Nat.Nat) |-
  n₁@0 = go₂ @0 (Where._⊗_ @1 @0) : Agda.Builtin.Nat.Nat
**********************


 -- Body -- 

Just (go₂ @0 (Where._⊗_ @1 @0))  ~~>  Just (go₂ (Where._⊗_ @1 @0))

 -- Child -- 

Where._._.go₂ :: (n₁ : Agda.Builtin.Nat.Nat) ->
  Agda.Builtin.Nat.Nat -> Agda.Builtin.Nat.Nat
Where._._.go₂ = [in range: /home/omelkonian/git/agda2hs/test/Where.agda:140,9-12 ]
[(n₁ : Agda.Builtin.Nat.Nat) (n₂ : Agda.Builtin.Nat.Nat) |-
   n₁@1 n₂@0 = Where._⊗_ (Where._⊗_ @2 @1) @0 : Agda.Builtin.Nat.Nat]

  ~~>(apply)  

 -- Child -- 

Where._._.go₂ :: Agda.Builtin.Nat.Nat -> Agda.Builtin.Nat.Nat
Where._._.go₂ = [in range: /home/omelkonian/git/agda2hs/test/Where.agda:140,9-12 ]
[(n₂ : Agda.Builtin.Nat.Nat) |-
   n₂@0 = Where._⊗_ (Where._⊗_ @1 @1) @0 : Agda.Builtin.Nat.Nat]

```

UlfNorell: I think the problem is that `apply`/`applyE` is not actually the right move here. What happens is this:
``(\ db1 db0 -> db2 * db1 * db0) `apply` db1`` --> `db1 * db1 * db0`. Basically, `apply` gets rid of a binder, thus shifting any outside de Bruijn indices down, but what you want in this case is to go inside the binder. So for definitions that would mean dropping the patterns for the outer variables in the clauses, and the corresponding arguments from the type.

omelkonian: Finally went with `apply`-ing and reverting all bodies (see `applyNoBodies` helper), kind of hack-ish but works :)

Main.hs
```diff
@@ -273,32 +288,86 @@ compileConstructorArgs (ExtendTel a tel)
...
+    let localUses = nub $ listify (`elem` map fst locals) body
+        belongs q@(QName m _) (QName m0 _) =
+          ((show m0 ++ "._") `isPrefixOf` show m) && (q `notElem` localUses)
+        splitDecls :: LocalDecls -> ([(Definition, LocalDecls)], LocalDecls)
+        splitDecls ds@((q,child):rest)
+          | q `elem` localUses
+          , (grandchildren, outer) <- span ((`belongs` q) . fst) rest
+          , (groups, rest') <- splitDecls outer
+          = ((child, grandchildren) : groups, rest')
+          | otherwise = ([], ds)
```
UlfNorell, line 345: This is too simplistic. If the first local decl is not in `localUses` it stops. Example:
```agda
ex8 : Nat
ex8 = n₂
  where
    n₁ = 1
    n₂ = n₁ + 1
```
compiles to
```haskell
ex8 :: Integer
ex8 = Where.n
```

Main.hs
```diff
@@ -273,32 +288,86 @@ compileConstructorArgs (ExtendTel a tel)
...
+        (children, locals') = splitDecls locals
+
+        args   = teleArgs tel
+        argLen = length args
+
+        -- 1. apply current telescope to inner definitions
+        children' = everywhere (mkT (`applyNoBodies` args)) children
+
+        -- 2. shrink calls to inner modules (unqualify + partially apply module parameters)
+        shrinkLocalDefs t | Def q es <- t, q `elem` concatMap (\(d,ds) -> defName d : map fst ds) children
```
UlfNorell, line 356: For clarity (and minor performance):
```suggestion
        localNames = concatMap (\ (d, ds) -> defName d : map fst ds) children
        shrinkLocalDefs t | Def q es <- t, q `elem` localNames
```

UlfNorell (changes requested):
We're getting there! Some cosmetics and a problem with finding local declarations.

Main.hs
omelkonian: Changing `elem` to `any elem` did the job.

UlfNorell (approved):
Good job!

PR #24, opened 2020-11-09
Framework for negative tests

UlfNorell:
My `make` skills are a bit rusty, so could probably be cleaned up a bit.


omelkonian (approved):
LGTM

PR #27, opened 2020-11-23
Fleshing out the Prelude

UlfNorell:
Still missing:
- Enum, Bounded
- Numeric stuff
- Partial functions and infinite lists
- Anything using Int
- Show, Read
- IO

### DIFF
--- a/lib/Haskell/Prelude.agda
+++ b/lib/Haskell/Prelude.agda
@@ -1,20 +1,76 @@
 {-# OPTIONS --no-auto-inline #-}
 module Haskell.Prelude where
 
-open import Agda.Builtin.Unit  public
-open import Agda.Builtin.List  public
-open import Agda.Builtin.Nat   public
-open import Agda.Builtin.Bool  public
-open import Agda.Builtin.Float public
-open import Agda.Builtin.Char  public
+open import Agda.Builtin.Unit         public
+open import Agda.Builtin.List         public
+open import Agda.Builtin.Nat   as Nat public hiding (_==_; _<_)
+open import Agda.Builtin.Bool         public
+open import Agda.Builtin.Float        public
+open import Agda.Builtin.Char         public
+open import Agda.Builtin.FromString   public
+import Agda.Builtin.String as Str
+open import Agda.Builtin.Strict
 
--- Type variables --
+-- Problematic features
+--  - [Partial]:  Could pass implicit/instance arguments to prove totality.
+--  - [Int]:      Agda doesn't have Int64. What to do?
+--  - [Float]:    Or Float (Agda floats are Doubles)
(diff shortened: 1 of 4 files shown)


lib/Haskell/Prelude.agda
```diff
@@ -1,20 +1,76 @@
...
+open import Agda.Builtin.Bool         public
+open import Agda.Builtin.Float        public
+open import Agda.Builtin.Char         public
+open import Agda.Builtin.FromString   public
+import Agda.Builtin.String as Str
+open import Agda.Builtin.Strict
 
--- Type variables --
+-- Problematic features
+--  - [Partial]:  Could pass implicit/instance arguments to prove totality.
```
omelkonian, line 15: This sounds reasonable, but we'll have to be careful to erase this totality information, and it will not suffice to just ignore implicit arguments for instance (since typeclasses in #3 use these). :thinking:

lib/Haskell/Prelude.agda
```diff
@@ -1,20 +1,76 @@
...
+--  - [Int]:      Agda doesn't have Int64. What to do?
+--  - [Float]:    Or Float (Agda floats are Doubles)
+--  - [Infinite]: Define colists and map to Haskell lists?
```
omelkonian, line 18: :+1:

lib/Haskell/Prelude.agda
```diff
@@ -1,20 +1,76 @@
...
+--     ShowS, Show(showsPrec, showList, show),
+--     shows, showChar, showString, showParen,
+--
+--     ReadS, Read(readsPrec, readList),
+--     reads, readParen, read, lex,
+--
+--     IO, putChar, putStr, putStrLn, print,
+--     getChar, getLine, getContents, interact,
+--     FilePath, readFile, writeFile, appendFile, readIO, readLn,
+--     IOError, ioError, userError,
```
omelkonian, line 62: Maybe add an issue/task for this?

omelkonian (approved):
Nice!

lib/Haskell/Prelude.agda
```diff
@@ -1,20 +1,76 @@
 {-# OPTIONS --no-auto-inline #-}
```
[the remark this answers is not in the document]
UlfNorell: Without auto-inlining Agda doesn't touch terms. We are currently reducing types ourselves, but I think we probably shouldn't.

Always turning off auto-inlining would make sense, but I'm not sure that the backend framework has a hook that lets us do that. FWIW auto-inlining is off by default in Agda 2.6.2.

lib/Haskell/Prelude.agda
UlfNorell: Yes, so they would have to go in implicit arguments rather than instance arguments.

PR #29, opened 2020-12-03
Numbers

UlfNorell:
Now maps
- `Agda.Builtin.Nat -> Numeric.Natural`
- `Agda.Builtin.Int -> Integer`
- `Agda.Builtin.Word64 -> Word` (instead of `Data.Word.Word64`)
- `Haskell.Prim.Int -> Int`

Also: special handling of `fromNat` and `fromNeg`.

Note: this means we are incompatible with 32-bit platforms (see #30)

### DIFF
--- a/HsUtils.hs
+++ b/HsUtils.hs
@@ -118,5 +118,5 @@ allUsedTypes = listify (const True)
 usedTypesOf :: Data a => String -> a -> [Type ()]
 usedTypesOf s = listify $ (== s) . pp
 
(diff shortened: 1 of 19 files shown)


jmchapman (approved):
It looks very nice! So, Agda Integers are mapped to Haskell Integers and you have modelled Int64 in Agda via Word64 that is mapped to Haskell's Int? Did you fix the 32-bit problem then?

UlfNorell:
No, this PR assumes 64-bit platform.

PR #32, opened 2020-12-07
Instance tweaks

UlfNorell:
Two things:
- Drop record argument also from functions defined in record modules (not just for projections). Useful for builtin classes which aren't going to be translated to Haskell. Functions that should be translated to Haskell cannot go in the record module at the moment, but possibly something could be done to allow this.
- Drop instance arguments that depend on visible arguments. This lets you have proof objects as instance arguments on the Agda side.

### DIFF
--- a/test/Test.agda
+++ b/test/Test.agda
@@ -141,6 +141,19 @@ sumMon []       = mempty
 sumMon (x ∷ xs) = x <> sumMon xs
 {-# COMPILE AGDA2HS sumMon #-}
 
+data NonEmpty {a : Set} : List a → Set where
+  instance
+    itsNonEmpty : ∀ {x xs} → NonEmpty (x ∷ xs)
+
+-- Instance argument proof obligation that should not turn into a class constraint
+hd : (xs : List a) → ⦃ NonEmpty xs ⦄ → a
+hd [] ⦃ () ⦄
+hd (x ∷ _) = x
+{-# COMPILE AGDA2HS hd #-}
+
+five : Int
+five = hd (5 ∷ 3 ∷ [])
+{-# COMPILE AGDA2HS five #-}
 
 -- ** Booleans
 
--- a/test/golden/Test.hs
+++ b/test/golden/Test.hs
@@ -75,6 +75,13 @@ sumMon :: Monoid a => [a] -> a
 sumMon [] = mempty
 sumMon (x : xs) = x <> sumMon xs
 
+hd :: [a] -> a
+hd [] = undefined
+hd (x₁ : xs) = x₁
+
+five :: Int
+five = hd (5 : 3 : [])
+
 ex_bool :: Bool
 ex_bool = True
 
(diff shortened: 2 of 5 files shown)


jmchapman:
Cool!
> * Drop record argument also from functions defined in record modules (not just for projections). Useful for builtin classes which aren't going to be translated to Haskell. Functions that should be translated to Haskell cannot go in the record module at the moment, but possibly something could be done to allow this.

Taking `Monoid` from `lib/Haskell/Prelude.hs` as an example:
```
record Monoid (a : Set) : Set where
  field
    mempty : a
    overlap ⦃ super ⦄ : Semigroup a

  mappend : a → a → a
  mappend = _<>_

  mconcat : List a → a
  mconcat []       = mempty
  mconcat (x ∷ xs) = x <> mconcat xs
```
Are you saying that now `mappend` will be considered a class member now but if cannot currently be compile to an ordinary function?

> Drop instance arguments that depend on visible arguments. This lets you have proof objects as instance arguments on the Agda side.

Nice!

test/Test.agda
```diff
@@ -139,6 +139,19 @@ sumMon []       = mempty
 sumMon (x ∷ xs) = x <> sumMon xs
 {-# COMPILE AGDA2HS sumMon #-}
 
+data NonEmpty {a : Set} : List a → Set where
+  instance
+    itsNonEmpty : ∀ {x xs} → NonEmpty (x ∷ xs)
+
+-- Instance argument proof obligation that should not turn into a class constraint
+hd : (xs : List a) → ⦃ NonEmpty xs ⦄ → a
```
jmchapman, line 147: ah, that works nicely :)

UlfNorell:
Yes exactly. Calls to `mappend` will work, but you couldn't put `mappendX` in the `MonoidX` record and have it compiled sensibly.

test/golden/Test.hs
```diff
@@ -71,6 +71,13 @@ sumMon :: Monoid a => [a] -> a
 sumMon [] = mempty
 sumMon (x : xs) = x <> sumMon xs
 
+hd :: [a] -> a
+hd [] = undefined
```
jmchapman, line 75: Interesting. Could we compile to `error` instead and get both totality and be more faithful to the Haskell code?

I guess one could have a custom eliminator for the empty type:
```
hd [] ⦃ () ⦄
```
```
hd [] ⦃ e ⦄ = error e "blah"
```

jmchapman (approved):

test/golden/Test.hs
UlfNorell: Yes, absurd clauses should probably compile to `error "impossible"` rather than `undefined`.

jmchapman: I think that would be nice and I guess even though uses of `hd` are guaranteed to be safe when writing Haskell code in Agda due to `NonEmpty` one can still use it unsafely from Haskell mode written directly in Haskell and trigger `error/undefined`.

PR #33, opened 2020-12-07
More Prelude

UlfNorell:
- `Show`
- `Num`
- `take`, `drop`, and `splitAt`
- Also changes the `Ord` class to include all the members from Haskell. It's not clear what the best design is here, but the problem I ran into was when you only have `compare` as the member and define `compare` using `<`, you get a very weird definition for the overloaded `<`. This way `<` maps to the actual function you used. An open question is how to compile instance declarations using the `ordFromX` helpers.

### DIFF
--- a/lib/Haskell/Prelude.agda
+++ b/lib/Haskell/Prelude.agda
@@ -2,7 +2,7 @@
 module Haskell.Prelude where
 
 open import Agda.Builtin.Unit         public
-open import Agda.Builtin.Nat as Nat   public hiding (_==_; _<_)
+open import Agda.Builtin.Nat as Nat   public hiding (_==_; _<_; _+_; _*_; _-_)
 open import Agda.Builtin.List         public
 open import Agda.Builtin.Bool         public
 open import Agda.Builtin.Float        public renaming (Float to Double)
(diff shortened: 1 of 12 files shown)


jmchapman (approved):
LGTM

jmchapman:
> * Also changes the `Ord` class to include all the members from Haskell. It's not clear what the best design is here, but the problem I ran into was when you only have `compare` as the member and define `compare` using `<`, you get a very weird definition for the overloaded `<`. This way `<` maps to the actual function you used.

Does this show up when you're reasoning about `<`? Or in some other way?

jmchapman:
> * Also changes the `Ord` class to include all the members from Haskell. It's not clear what the best design is here, but the problem I ran into was when you only have `compare` as the member and define `compare` using `<`, you get a very weird definition for the overloaded `<`. This way `<` maps to the actual function you used.

Does this show up when you're reasoning about `<`? Or in some other way?

UlfNorell:
Yes, for reasoning. In the `Num` instance for `Nat` I put a constraint `IsFalse (n < m)` for subtraction, which instead of turning into `n Nat.< m` became a monstrosity like `\ n m -> (if n Nat.< m then LT else if n Nat.== m then EQ else GT) == LT`.

I think currently it's going directly for `Nat._<_`, but with the refactor `Ord` they are the same.

jmchapman:
That does look especially monstrous!

PR #36, opened 2020-12-08
List literals

UlfNorell:
(on top of #35)

### DIFF
--- a/lib/Haskell/Prelude.agda
+++ b/lib/Haskell/Prelude.agda
@@ -17,7 +17,7 @@ open import Agda.Builtin.Equality     public
 open import Agda.Builtin.Int using (pos; negsuc)
 
 open import Haskell.Prim
-open Haskell.Prim public using (TypeError; ⊥; if_then_else_; iNumberNat; IsTrue; IsFalse; All; allNil; allCons)
+open Haskell.Prim public using (TypeError; ⊥; if_then_else_; iNumberNat; IsTrue; IsFalse; All; allNil; allCons; NonEmpty)
 
 open import Haskell.Prim.Integer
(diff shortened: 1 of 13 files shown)


omelkonian (approved):
Much more idiomatic, indeed! 

Another way to go about it is support both, by providing appropriate syntax on the Agda side.
For example, you can do [this trick](https://github.com/omelkonian/formal-prelude/blob/master/Prelude/Nary.agda#L8-L13) and the users will be able to choose whatever they prefer.

UlfNorell:
Only on the right-hand side though.

lib/Haskell/Prelude.agda
```diff
@@ -116,6 +125,16 @@ case x of f = f x
 asTypeOf : a → a → a
 asTypeOf x _ = x
 
+undefined : {@(tactic absurd) i : ⊥} → a
+undefined {i = ()}
+
+error : {@(tactic absurd) i : ⊥} → String → a
```
jmchapman, line 131: this is really cool!

test/Test.agda
```diff
@@ -159,13 +163,9 @@ sumMon []       = mempty
...
 {-# COMPILE AGDA2HS sumMon #-}
 
-data NonEmpty {a : Set} : List a → Set where
-  instance
-    itsNonEmpty : ∀ {x xs} → NonEmpty (x ∷ xs)
-
 -- Instance argument proof obligation that should not turn into a class constraint
 hd : (xs : List a) → ⦃ NonEmpty xs ⦄ → a
-hd [] ⦃ () ⦄
+hd []      = error "hd: empty list"
```
jmchapman, line 168: this is very cool - best of both worlds.

Issue #42, opened 2020-12-10
deriving


jmchapman:

UlfNorell:
Possible design:
```agda
data MyDatatype : Set where
  ...
{-# COMPILE AGDA2HS MyDatatype deriving (Show, Eq, Ord) #-}
```

jespercockx:
Wouldn't you also want the derived instances be available on the Agda side? Which would mean you'd have to implement some tactics for automatically deriving instances in Agda.

omelkonian:
Let's assume you have a deriving macro (I already have something [here](https://github.com/omelkonian/formal-prelude/blob/a3ff049c5d5af002efdac8b2e903e34d0ad98a6e/Prelude/Generics.agda#L185-L199), e.g. for [DecEq](https://github.com/omelkonian/formal-prelude/blob/a3ff049c5d5af002efdac8b2e903e34d0ad98a6e/Prelude/DecEq.agda#L270)).
It can be used like this:
```agda
data ℕ′ : Set where
  O : ℕ′
  S : ℕ′ → ℕ′
unquoteDecl DecEq-ℕ′ = DERIVE DecEq [ quote ℕ′ , DecEq-ℕ′ ]
```
My question is, will we be able to grab the DERIVE directive before unquoting, so as to emit Haskell deriving alongside the datatype? Does the current Agda backend infra support such a thing?

jespercockx:
I believe it's currently not possible for a backend to determine whether a function was written using `unquoteDecl` or by hand, unfortunately. So the declaration would be translated to a hand-written instance in Haskell. Maybe that's anyway safer, since I don't see how you would guarantee that the Agda implementation of the instance corresponds to the one derived by Haskell.

UlfNorell:
It's certainly safer to translate all code from Agda, but I think in many cases you are not super worried about getting the instance wrong and then it would be convenient to only define it on the Agda side and have it derived on the Haskell side. Whether the Agda definition is done with a macro or by hand doesn't really make a difference.

jmchapman:
I think if one uses `deriving` in it's simplest form one should postulate the functionality in Agda (not sure how this works with instances though) following the usual guidance described in the [manual](https://agda.readthedocs.io/en/latest/language/foreign-function-interface.html#using-haskell-functions-from-agda). I think it's analogous to using code written by a third party from an external haskell module (e.g. bytetring) except here the third party is ghc.

[...]

PR #46, opened 2020-12-14
added support for deriving in a pragma

jmchapman:
I followed @UlfNorell's suggestion https://github.com/agda/agda2hs/issues/42#issuecomment-742502084

Eg.,
```
data Exp (v : Set) : Set where
  Plus : Exp v → Exp v → Exp v
  Lit : Nat → Exp v
  Var : v → Exp v
{-# COMPILE AGDA2HS Exp deriving (Show,Eq) #-}
```
->
```
data Exp v = Plus (Exp v) (Exp v)
           | Lit Natural
           | Var v
               deriving (Show, Eq)
```

Fixes https://github.com/agda/agda2hs/issues/42

### DIFF
--- a/Main.hs
+++ b/Main.hs
@@ -344,7 +344,7 @@ processPragma qn = getUniqueCompilerPragma pragmaName qn >>= \case
     case Hs.parseDecl ("data X = X " ++ s) of
       Hs.ParseFailed loc msg ->
         setCurrentRange (srcLocToRange loc) $ genericError msg
-      Hs.ParseOk m | Hs.DataDecl _ _ _ _ _ ds <- m ->
+      Hs.ParseOk (Hs.DataDecl _ _ _ _ _ ds) ->
         return $ DerivingPragma ds
       Hs.ParseOk _ -> __IMPOSSIBLE__
   _ -> return DefaultPragma
--- a/Main.hs
+++ b/Main.hs
@@ -320,18 +320,22 @@ tuplePat cons i ps = do
 
 data CodeGen = YesCode | NoCode
 
-data ParsedResult
+data ParsedPragma
   = NoPragma
   | DefaultPragma
   | ClassPragma CodeGen
-  | DerivingPragma [Hs.Deriving Hs.SrcSpanInfo]
+  | DerivingPragma [Hs.Deriving ()]
 
 classes :: [String]
 classes = ["Show"]
 
 -- "class" is not being used usefully, any record with a pragma is
 -- considered a typeclass
-processPragma :: QName -> TCM ParsedResult
+
+-- no pragma at all means no code is compiled
+-- if the pragma contains extraneous stuff we treat it as default
+-- using a class pragma currently leads to no code being compiled
+processPragma :: QName -> TCM ParsedPragma
 processPragma qn = getUniqueCompilerPragma pragmaName qn >>= \case
   Nothing -> return NoPragma
   Just (CompilerPragma _ s) | s == "class" && s `elem` classes ->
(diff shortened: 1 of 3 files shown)


Main.hs
```diff
@@ -337,22 +338,33 @@ processPragma qn = getUniqueCompilerPragma pragmaName qn >>= \case
...
     return $ ClassPragma YesCode
-  _                                        -> return DefaultPragma
-
+  Just (CompilerPragma _ s) | "deriving" `isPrefixOf` s ->
+    -- parse a deriving clause for a datatype by tacking it onto a
+    -- dummy datatype and then only keeping the deriving part
+    case Hs.parseDecl ("data X = X " ++ s) of
+      Hs.ParseFailed loc msg ->
+        setCurrentRange (srcLocToRange loc) $ genericError msg
+      Hs.ParseOk m | Hs.DataDecl _ _ _ _ _ ds <- m ->
```
UlfNorell, line 347:
```suggestion
      Hs.ParseOk (Hs.DataDecl _ _ _ _ _ ds) ->
```

Main.hs
```diff
@@ -337,22 +338,33 @@ processPragma qn = getUniqueCompilerPragma pragmaName qn >>= \case
...
+        return $ DerivingPragma ds
+      Hs.ParseOk _ -> __IMPOSSIBLE__
+  _ -> return DefaultPragma
 compile :: Options -> ModuleEnv -> IsMain -> Definition -> TCM CompiledDef
 compile _ _ _ def = processPragma (defName def) >>= \ case
   NoPragma -> return []
   DefaultPragma -> compile' def
   ClassPragma NoCode -> return []
   ClassPragma YesCode -> return []
+  DerivingPragma ds   -> tag <$> compileData (map (() <$) ds) def
```
UlfNorell, line 357: We haven't checked that the def is a `Datatype` here, so will hit the `__IMPOSSIBLE__` in `compileData` if you put `deriving` on a non-datatype. I wonder if a nicer solution would be to pass the pragma to `compile'`. That would also get rid of the duplication of `tag`.

UlfNorell (approved):

UlfNorell:
Don't forget to put a "fixes #42" in the description so it auto closes the issue when merged.

Main.hs
```diff
@@ -337,22 +338,33 @@ processPragma qn = getUniqueCompilerPragma pragmaName qn >>= \case
...
+      where tag code = [(nameBindingSite $ qnameName $ defName def, code)]
```
omelkonian, line 358: I think it makes sense to merge `compile` with `compile'`, I don't see any use-sites that uses only one of them.

Main.hs
```diff
@@ -337,22 +338,33 @@ processPragma qn = getUniqueCompilerPragma pragmaName qn >>= \case
     return $ ClassPragma NoCode
   Just (CompilerPragma _ s) | s == "class" ->
     return $ ClassPragma YesCode
-  _                                        -> return DefaultPragma
-
+  Just (CompilerPragma _ s) | "deriving" `isPrefixOf` s ->
+    -- parse a deriving clause for a datatype by tacking it onto a
+    -- dummy datatype and then only keeping the deriving part
+    case Hs.parseDecl ("data X = X " ++ s) of
```
omelkonian, line 344: That's quite a hack, but it seems src-exts does not provide its internal Happy parsers readily, so I guess this is fine.

omelkonian (approved):
Nice!

Main.hs
jmchapman: Thanks for spotting that! I have merged `compile` and `compile'`.

Main.hs
jmchapman: I have merged `compile` and `compile'`.

Main.hs
jmchapman: Yes :) It seemed the easiest way to do it as `Language.Haskell.Exts.Parser` parser only seems to have functions/instances for bigger lumps of AST.

PR #47, opened 2020-12-15
Refactor: Split up the Prelude in smaller Prim modules

UlfNorell:


jmchapman (approved):
Nice!
Seems sensible to structure it rather than having one giant file and the renamings of dropping the `64` from `Word64` and `Int64` sounds good too.

PR #49, opened 2021-01-18
Add Enum and Bounded to Prelude

UlfNorell:
... and compile to enumFromTo sugar.

Not quite perfect: with the current representation an `Enum` is either `Bounded` or unbounded, which works for most types, but notably not for `Natural` which is only bounded below.

Possible solution would be to split `Bounded` into `BoundedAbove` and `BoundedBelow` and base the `Enum` constraints on these.

### DIFF
--- a/lib/Haskell/Prim/Enum.agda
+++ b/lib/Haskell/Prim/Enum.agda
@@ -182,7 +182,7 @@ module _ (from : a → Integer) (to : Integer → a) where
 
 instance
   iEnumNatural : Enum Nat
-  iEnumNatural = boundedBelowEnumViaInteger pos λ where (pos n) → n; _ → 0
+  iEnumNatural = boundedBelowEnumViaInteger pos unsafeIntegerToNat
 
   iEnumInt : Enum Int
   iEnumInt = boundedEnumViaInteger intToInteger integerToInt
--- a/lib/Haskell/Prim/Enum.agda
+++ b/lib/Haskell/Prim/Enum.agda
@@ -1,0 +1,168 @@
+
+module Haskell.Prim.Enum where
+
+open import Agda.Builtin.Nat as Nat hiding (_==_; _<_; _+_; _*_; _-_)
+open import Agda.Builtin.Char
+open import Agda.Builtin.Equality
+open import Agda.Builtin.List
+open import Agda.Builtin.Unit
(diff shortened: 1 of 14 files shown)


jmchapman:
I think I may not have grasped the point here!

I think `Natural` is not an instance of `Bounded` in Haskell (https://hackage.haskell.org/package/base-4.14.1.0/docs/Numeric-Natural.html). So, are we already in the same position? Are you then suggesting to introduce more refined versions of `Bounded` (i.e., `BoundedAbove` and `BoundedBelow`) to have something nicer in Agda? I also read that for any "enumeration type" one can derive `Bounded` but I take this to mean a non-recursive data type rather than an instance of class `Enum`.

UlfNorell:
Yes the suggestion was to split `Bounded` in two classes `BoundedBelow` and `BoundedAbove` and define `Bounded` as simply wrapping those two. We could then give an instance `BoundedBelow Natural`.

Instead of a `Maybe Bounded` in the `Enum` class, we could have one `Maybe BoundedBelow` and one `Maybe BoundedAbove` and express the method constraints in terms of those.

jmchapman:
Thanks for clarifying. That sounds good to me.

UlfNorell:
Now with `Enum Natural` instance

lib/Haskell/Prim/Enum.agda
```diff
@@ -0,0 +1,208 @@
...
+  boundedEnumViaInteger .enumFromTo     a b    = fromTo a b
+  boundedEnumViaInteger .enumFromThenTo a a₁ b = fromThenTo a a₁ b
+  boundedEnumViaInteger .enumFrom       a      = fromTo a maxBound
+  boundedEnumViaInteger .enumFromThen   a a₁   =
+    if a < a₁ then fromThenTo a a₁ maxBound
+              else fromThenTo a a₁ minBound
+
+instance
+  iEnumNatural : Enum Nat
+  iEnumNatural = boundedBelowEnumViaInteger pos λ where (pos n) → n; _ → 0
```
omelkonian, line 185: The λ-expression is just your previously defined `unsafeIntegerToNat`.

omelkonian (approved):
LGTM, just some minor suggestions.

UlfNorell:
All good suggestions. 👍

jmchapman:
Nice!

PR #53, opened 2021-04-20
Issue 51

### LINKED ISSUE #51: Defining custom typeclass instances
I'm trying to use `agda2hs` and having trouble understanding how to define your own record instances and export them properly.

My  understanding is that the Haskell prelude from `agda2hs` is made so that it provides the same interface as the Haskell prelude, such that when you use it inside an Agda definition, and compile the definition to Haskell, it will use the Haskell prelude directly.

However if I want to define, let's say, a `Monoid` instance for my own Agda datatype, and compile it to Haskell code, using the `Monoid` record defined by `agda2hs` doesn't work, as it doesn't map 1-to-1 to the Haskell definition (notably, it has a `super` field).

That is, the following Agda code:
```agda
instance
  listMonoid : ∀ {a} → Monoid (List a)
  listMonoid {a} .mempty = Nil
{-# COMPILE AGDA2HS listMonoid #-}
```
will produce the following Haskell code:
```hs
instance Monoid (List a) where
        mempty = Nil
        super = listSemigroup
```

There seem to be two solutions:
- Don't export the `listMonoid` instance and write by hand the Haskell one:
  ```agda
  {-# FOREIGN AGDA2HS
    instance Monoid (List a) where
      mempty  = Nil
      mappend = concat
    #-}
  ```
  That's unfortunate as it introduces code duplication.
- Ignore the `agda2hs` definition of `Monoid` and use one that actually is identical to the one in the Haskell prelude.

Thus my questions:
- Why wasn't the `Monoid`, `Monad` records (and more) not made identical to their Haskell counterpart?
- Is there an idiomatic way to handle this discrepancy?

UlfNorell:
When generating instance declarations, only explicit fields should be used as members. Implicit and instances fields are used for additional proofs, or super class constraints, so should be dropped from the Haskell code.

Fixes #51

### DIFF
--- a/Main.hs
+++ b/Main.hs
@@ -398,7 +398,7 @@ compileInstance def = do
   locals <- takeWhile (isAnonymousModuleName . qnameModule . fst)
           . dropWhile ((<= defName def) . fst)
           . sortDefs <$> curDefs
-  ds <- concat <$> mapM (compileInstanceClause locals) funClauses
+  ds <- catMaybes <$> mapM (compileInstanceClause locals) funClauses
   return $ [Hs.InstDecl () Nothing ir (Just ds)]
   where Function{..} = theDef def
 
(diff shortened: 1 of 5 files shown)


Main.hs
```diff
@@ -425,7 +425,7 @@ compileInstRule cs ty = case unSpine  $ ty of
     where dropPi = underAbstr a b (compileInstRule cs . unEl)
   _ -> __IMPOSSIBLE__
 
-compileInstanceClause :: LocalDecls -> Clause -> TCM (Hs.InstDecl ())
+compileInstanceClause :: LocalDecls -> Clause -> TCM [Hs.InstDecl ()]
```
omelkonian, line 428: Since we are returning either 0 or 1 declarations (depending on the visibility of the argument), I would prefer returning `TCM (Maybe (Hs.InstDecl ()))`, and then using `catMaybe` instead of `concat` in `compileInstance`.

omelkonian (approved):

PR #61, opened 2021-04-30
[ fix #59 ] Add translation to Haskell records

### LINKED ISSUE #59: Compiling records to Haskell single-constructor datatypes
In Haskell it seems frequent to emulate "record types" by defining single-constructor datatypes with named fields. When defined as such, they behave much like agda's record types.

Right now, it looks like agda2hs compiles all record types into class definitions.
I was wondering whether it would make sense to let the user compile some record types to Haskell "records" as such:

```agda
record QuadTree (a : Set) : Set where
  constructor Wrapper
  field
    wrappedTree : Quadrant a
    treeLength  : Int
    treeWidth   : Int
    treeDepth   : Int
{-# COMPILE AGDA2HS record QuadTree #-}
```
into
```hs
data QuadTree a = Wrapper { wrappedTree :: Quadrant a
                          , treeLength  :: Int
                          , treeWidth   :: Int
                          , treeDepth   :: Int
                          }
```
I'd be willing to give a try at implementing this, if it is indeed a desirable feature.

flupe:
Given `Rec` an agda record type, `{-# COMPILE AGDA2HS Rec #-}` will now produce an Haskell record type, and `{-# COMPILE AGDA2HS Rec class #-}` will produce a class definition.
Support for `deriving` could most likely be added later.

### DIFF
--- a/Main.hs
+++ b/Main.hs
@@ -343,12 +343,12 @@ tuplePat cons i ps = do
 
 -- Compiling things -------------------------------------------------------
 
-data CodeGen = YesCode | NoCode
+data RecordTarget = ToRecord | ToClass
 
 data ParsedPragma
   = NoPragma
   | DefaultPragma
-  | ClassPragma CodeGen
+  | ClassPragma
   | DerivingPragma [Hs.Deriving ()]
 
 classes :: [String]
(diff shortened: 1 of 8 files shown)


Main.hs
```diff
@@ -457,36 +454,53 @@ fieldArgInfo f = do
...
+      classDecls <- compileRecFields classDecl recPars (unDom <$> recFields) recTel
+      return [Hs.ClassDecl () Nothing hd [] (Just classDecls)]
+
+    ToRecord | recNamedCon -> do
+      fieldDecls <- compileRecFields fieldDecl recPars (unDom <$> recFields) recTel
+      let conDecl = Hs.QualConDecl () Nothing Nothing $ Hs.RecDecl () cName fieldDecls
+      return [Hs.DataDecl () (Hs.DataType ()) Nothing hd [conDecl] []]
+
+             | otherwise   -> genericError $
+      "Not supported: " ++ rName ++ " is missing a named constructor."
```
UlfNorell, line 477: You need to `setCurrentRange` here to get a source location on the error, but I can fix that post-merge.

UlfNorell (approved):
Very nice. Thank you.

PR #66, opened 2021-05-24
Drop non-visible lambdas

### LINKED ISSUE #65: Instance arguments failed to compile
Instance arguments sometimes fail to compile, for example, this fails
```agda
module Data.QuadTree.Test where

open import Haskell.Prelude 

yeet : {a : Set } → (c : Bool) → ({{Bool}} → a) → a
yeet false x = x {{false}}
yeet true x = x {{true}}
{-# COMPILE AGDA2HS yeet #-}

xx = yeet true false
{-# COMPILE AGDA2HS xx #-}
```

With the error:
```
/home/jonathan/projects/research-project/src/Data/QuadTree/Test.agda:10,1-3
bad term: λ ⦃ _ ⦄ → false
```

I was not able to reduce it any further, however, I have found a modification to agda2hs that fixes this problem. This modification probably allows way too much though, so would be nice if someone else could make a proper PR

Adding the following 3 lines fixes the problem:
https://github.com/JonathanBrouwer/agda2hs/blob/3486a745b7e7f681b6f8ff1577db470c33216a18/Main.hs#L755-L757

UlfNorell:
Fixes #65

Note that the test case in the OP doesn't work since `{{Bool}} -> a` gets compiled to `Bool => a`. Making it a proper predicate on `c` makes the compiler drop it though.

### DIFF
--- a/Main.hs
+++ b/Main.hs
@@ -768,6 +768,9 @@ compileTerm v =
           Hs.InfixApp _ a op b
             | a == hsx -> Hs.RightSection () op b -- System-inserted visible lambdas can only come from sections
           _            -> hsLambda x body         -- so we know x is not free in b.
+    Lam v b ->
+      -- Drop non-visible lambdas (#65)
+      underAbstraction_ b $ \ body -> compileTerm body
     t -> genericDocError =<< text "bad term:" <?> prettyTCM t
   where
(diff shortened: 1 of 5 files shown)


omelkonian:
LGTM!

PR #70, opened 2021-06-08
[ fix #69 ] Compile constructor args in the same way as function args

### LINKED ISSUE #69: Constructor with instance argument not compiled correctly (?)
Here's an example (by one of my students) of a datatype that is currently not handled correctly:
```agda
mutual

  data Map (k : Set) (a : Set) : Set where
    Bin : (sz : Nat) → (kx : k) → (x : a)
          → (l : Map k a) → (r : Map k a)
          → {{szVal : sz ≡ (size l) + (size r) + 1}}
          → Map k a
    Tip : Map k a
  {-# COMPILE AGDA2HS Map #-}

  size : {k a : Set} → Map k a → Nat
  size Tip = 0
  size (Bin sz _ _ _ _) = sz
  {-# COMPILE AGDA2HS size #-}
```

jespercockx:


omelkonian (approved):
LGTM

Issue #73, opened 2021-06-15
Implicit fields in record types are being compiled to Haskell


RicovanBuren:
Currently implicit fields in record types are being compiled to Haskell by agda2hs. This causes problems when the implicit field cannot be compiled to Haskell. Normally, implicit arguments are being ignored by agda2hs, I think this should also be the case for implicit fields. See the code below for an example:
```Agda
module ImplicitFieldTest where

record ImplicitField (a : Set) : Set where
    field
        aField : a
        {anImplicitField} : a
{-# COMPILE AGDA2HS ImplicitField #-}
```

Compiles to:
```Haskell
module ImplicitFieldTest where

class ImplicitField a where
    aField :: a
    anImplicitField :: a
```

PR #76, opened 2021-06-18
Fix bug in Eq instance for List

dxts:
https://github.com/agda/agda2hs/blob/1b350f85ed02f44dd24c07bf7f147055cac6da2e/lib/Haskell/Prim/Eq.agda#L67-L70

Pretty sure it should be
```
iEqList ._==_  []  [] = true
```


UlfNorell:
Oops

gallais:
Could laws be included as a bunch of `@0` field? It'd even be possible to
postulate function extensionality to prove them given that the proofs will
never end up in the compiled code!

UlfNorell:
I guess they could, but it would mean you would always have to prove them. Also they wouldn't even have to be `@0` since agda2hs erases all proofs (or is supposed to: #73).

PR #78, opened 2021-07-07
Agda-2.6.2 compatibility

### LINKED ISSUE #79: Agda-2.6.2
Two issues:
- The internals have changed in 2.6.2. The most annoying change being no `Data` instance for `Definition`
- `Agda.Builtin.Reflection` has changed, breaking the `Haskell.Prim.Absurd` tactic.

The first issue is easily solved with some `#if` statements. See #78. Not sure what the best way to handle the second issue is. We could
- drop 2.6.1 support
- split the lib in `lib-2.6.1` and `lib-2.6.2` (lots of duplication and not so nice for maintenance)
- ??

UlfNorell:
Only for the implementation, not for the agda lib.

Fixes #79

- ~~[ ] CI for both 2.6.1 and 2.6.2~~
- ~~[ ] Figure out how to handle the Agda library~~

Edit: as discussed in #79, we're dropping support for 2.6.1.

If anyone desperately needs it in the future, it should be easy enough to make a separate branch for it.

### DIFF
--- a/Main.hs
+++ b/Main.hs
@@ -2,11 +2,13 @@
 module Main where
 
 import Control.Applicative
-import Control.Arrow ((>>>), (***))
+import Control.Arrow ((>>>), (***), (&&&), first, second)
 import Control.Monad
 import Control.Monad.Fail (MonadFail)
 import Control.Monad.Reader
 import Control.Monad.IO.Class
+import Control.DeepSeq
+import Data.Data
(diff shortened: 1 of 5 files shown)


jespercockx:
Can this be merged now?

UlfNorell:
Let's hope so, because I just did.

PR #96, opened 2022-05-12
[ fix #95 ] Define tuples by large elimination

### LINKED ISSUE #95: Tuples are not recognized as strictly-positive in their arguments
Agda's positivity checker doesn't recognize tuples (as implemented in `agda2hs`, i.e heterogeneous lists indexed by a list of types) as strictly-positive.

```agda
open import Haskell.Prelude

data Test : Set where
  test : Test × String → Test
```

Gives the following error:

```
Test is not strictly positive, because it occurs
in the first argument of Tuple
in the type of the constructor test
in the definition of Test.
```

flupe:
Might have to make `Pair` *private* (in `Prim/Tuple.agda`) to avoid things going wrong.
Other than that this fixes the original issue.

A nice byproduct is that now `Tuple []` and `⊤` are the same.

### DIFF
--- a/Main.hs
+++ b/Main.hs
@@ -249,19 +249,19 @@ isSpecialTerm q = case prettyShow q of
 
 isSpecialCon :: QName -> Maybe (ConHead -> ConInfo -> Elims -> C (Hs.Exp ()))
 isSpecialCon = prettyShow >>> \ case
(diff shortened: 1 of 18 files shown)


jespercockx:
I brought this up to date with current master in #102, let's continue the discussion there.

Issue #93, opened 2022-02-25
`where` clause following `case_of_ λ where` does not compile


flupe:
For some reason, adding a `where` clause below a `case_of_ λ where` does not compile, failing with:

    Pattern matching lambdas must take a single argument

This fails:

```agda

open import Haskell.Prelude

fun : Bool → Bool
fun x = case x of λ where
          True  → False
          False → y
  where
    y : Bool
    y = True
{-# COMPILE AGDA2HS fun #-}
```

This works fine:

```agda

open import Haskell.Prelude

fun : Bool → Bool
fun x = impl
  where
    y : Bool
    y = True

    impl : Bool
    impl = case x of λ where
             True  → False
             False → y
{-# COMPILE AGDA2HS fun #-}
```

Issue #92, opened 2022-02-24
Ill-formed type signature in nested where block


jespercockx:
Minimal example:
```agda
open import Haskell.Prelude

postulate Something : Set
postulate something : Something

foo : a → a
foo {a} x = bar {something}
  where
    bar : @0 {Something} → a
    bar {eq} = baz
      where
        baz : a
        baz = x
{-# COMPILE AGDA2HS foo #-}
```
Compiles to the following:
```agda
foo :: a -> a
foo x = bar
  where
    bar :: a
    bar = baz
      where
        baz :: x
        baz = x
```
Note that `baz :: x` instead of `baz :: a` in the compiled code.

PR #97, opened 2022-05-13
refactoring where-clauses

### LINKED ISSUE #90: Definition in where block is not compiled
Consider the following example:
```agda
open import Haskell.Prelude

myNat : Nat
myNat = bar
  where
    bar : Nat
    foo : Nat
    foo = 42
    bar = foo
{-# COMPILE AGDA2HS myNat #-}
```
Agda2hs compiles this to the following:
```haskell

myNat :: Natural
myNat = bar
  where
    bar :: Natural
    bar = foo
```
Note that the definition of `foo` has disappeared. The problem does not occur when the type signature for `bar` is given right before its definition. 

I tried looking at the relevant part of the code in Agda2Hs but it is rather inscrutable. @UlfNorell could you perhaps take a look?

omelkonian:
* Fixes #90,#92,#93
    (no test case for #92 yet, needs explicit `forall`)

  * Utilize the Reader monad of `C` to track local where-declarations.

  * More modular code structure, to make it easier to generalize for
    other backends later on.


jespercockx:
I've brought this up-to-date with current master, I think it's good to merge this soon in order to avoid further problems. @omelkonian are there any changes you still wish to make or are you fine with merging this as-is?

jespercockx:
I refactored this fix a bit so it makes more reuse of some of the Agda internals, and it now also is more lazy with dropping arguments corresponding to module parameters.

jespercockx:
I'm going ahead and merge this now, we can always improve things further at a later point.

omelkonian:
> @omelkonian are there any changes you still wish to make or are you fine with merging this as-is?

I was hoping to make this much simpler by tracking provenance of `where` clauses [in Agda itself](https://github.com/agda/agda/pull/5906), but it turned out we still need the ugly tracking-down below due to pattern lambdas:
```hs
-- | Before checking a function, grab all of its local declarations.
-- TODO: simplify this when Agda exposes where-provenance in 'Internal' syntax
withFunctionLocals :: QName -> C a -> C a
withFunctionLocals q k = do
  ls <- takeWhile (isAnonymousModuleName . qnameModule)
      . dropWhile (<= q)
      . map fst
      . sortDefs <$> liftTCM curDefs
  withLocals ls k
```

I hope we can find a way to simplify this even more in the future, and then port the same behaviour to MAlonzo.

PR #101, opened 2022-08-03
Add support for do-notation

jespercockx:
This is a first attempt at adding support for do-notation to Agda2Hs. There are still two problems with this implementation:

- It does not detect all uses of do-notation in Agda, see the test case. The reason is that Agda tries to eta-contract the second argument of `>>=`, but I currently rely on the presence of a lambda to turn `>>=` back into do-notation.
- Conversely, the current implementation will turn *all* occurrences of `>>=` where the second argument is a lambda into `do`-notation, even when the user didn't use do-notation in the Agda source.

Both problems could be prevented if only Agda would keep track of a bit more information, in particular about the origin of each `QName`: whether it was written by the user, generated by desugaring (such as for `do`-notation) or system-generated. Without this change to Agda, it seems very difficult to build better support for do-notation.

### DIFF
--- a/lib/Haskell/Prim/Monad.agda
+++ b/lib/Haskell/Prim/Monad.agda
@@ -29,6 +29,15 @@ record Monad (m : Set → Set) : Set₁ where
   _=<<_ : (a → m b) → m a → m b
   _=<<_ = flip _>>=_
 
+-- Use `Dont._>>=_` and `Dont._>>_` if you do not want agda2hs to use
+-- do-notation.
+module Dont where
+  _>>=_ : ⦃ Monad m ⦄ → m a → (a → m b) → m b
+  _>>=_ = Monad._>>=_ it
+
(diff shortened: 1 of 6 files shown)


jespercockx:
@flupe: We could solve the issues here by always translating `_>>=_` to do-notation, but also provide a module (which is not opened) that contains a synonym for `_>>=_` that is *not* translated to do-notation, which you can use in a qualified way.

jespercockx:
I've implemented the approach suggested by @flupe. Any objections to merging this (@omelkonian @jmchapman)?

jmchapman:
This sounds good but would it be possible to do it the other way around? i.e., translate do to a qualified version of >>= and the translate that back to do. And if one uses >>=  then leave it as is.

jespercockx:
After discussing with @jmchapman and @omelkonian we decided to keep the current behaviour in this PR.

PR #102, opened 2022-08-03
Define tuples by large elimination (cont)

jespercockx:
Continuation of #96


jespercockx:
One disadvantage of defining tuples by large elimination is that it makes instance search less powerful on the side of Agda, i.e. Agda might think that there is overlap between instances in some cases. I don't think it's a big problem in this case as long as your instances follow the general pattern of having an instance for the empty tuple and one for the extended tuple, but in general it's something to be aware of. Probably this is still worth merging.

PR #104, opened 2022-09-16
[ new ] Add support for transparent function that are inlined in Haskell code

jespercockx:
This allows us to get rid of function calls that are simply the identity after erasure, e.g. `subst`.

### DIFF
--- a/src/Agda2Hs/Compile.hs
+++ b/src/Agda2Hs/Compile.hs
@@ -7,7 +7,7 @@ import Agda.TypeChecking.Pretty
 
 import Agda2Hs.Compile.ClassInstance ( compileInstance )
 import Agda2Hs.Compile.Data ( compileData )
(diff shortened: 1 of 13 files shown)


gallais:
There's already a (more powerful) `isIdentity` function in `Agda.Compiler.Treeless.Identity`.
Is the issue that y'all are working at a level above treeless? Is it not worth producing the treeless
representation to automatically detect the (potentially recursive) identity functions?

jespercockx:
The criterion for correctness here is really that the extracted Haskell function is an identity function. It is a bit unfortunate that we have to duplicate the functionality, but it feels to me that this is the most robust/reliable way to do it.

jespercockx:
Also, I just pushed some updates to make the check more general, so it now also handles recursive identity functions.

PR #111, opened 2022-10-03
Compatibility with --erased-cubical

jespercockx:
It would be cool if we could use `--erased-cubical` together with Agda2Hs so we can use cubical features to prove things such as functional extensionality and that bisimilarity implies equality for coinductive data structures. It would also enable the use of (erased) higher constructors to enforce that functions on a given datatype always respect certain semantic equivalences.

This PR makes three relatively small changes to enable this:
- Remove erasure annotations from levels and types (see https://github.com/agda/agda/issues/6114)
- Rewrite a few functions in the prelude to make them pass the stricter `--without-K` termination checker
- Make arguments of type `Size` implicit so they are erased even without a `@0` annotation (basically giving them the same treatment as levels)

### DIFF
--- a/agda2hs.agda-lib
+++ b/agda2hs.agda-lib
@@ -1,3 +1,4 @@
 name: agda2hs
 depend:
 include: lib
+flags: --without-K
--- a/agda2hs.agda-lib
+++ b/agda2hs.agda-lib
@@ -1,4 +1,3 @@
 name: agda2hs
 depend:
 include: lib
-flags: --erase-record-parameters
--- a/lib/Haskell/Prim/Eq.agda
+++ b/lib/Haskell/Prim/Eq.agda
@@ -60,9 +60,13 @@ instance
   iEqTuple ._==_ (x ; xs) (y ; ys) = x == y && xs == ys
 
   iEqList : ⦃ Eq a ⦄ → Eq (List a)
-  iEqList ._==_ []       []       = True
-  iEqList ._==_ (x ∷ xs) (y ∷ ys) = x == y && xs == ys
-  iEqList ._==_ _        _        = False
+  iEqList {a} ._==_ = eqList
+    where
+      eqList : List a → List a → Bool
+      eqList [] [] = True
+      eqList (x ∷ xs) (y ∷ ys) = x == y && eqList xs ys
+      eqList _ _ = False
+
 
   iEqMaybe : ⦃ Eq a ⦄ → Eq (Maybe a)
   iEqMaybe ._==_ Nothing  Nothing  = True
--- a/lib/Haskell/Prim/Thunk.agda
+++ b/lib/Haskell/Prim/Thunk.agda
@@ -6,7 +6,7 @@ open import Agda.Builtin.Size public
 
 open import Haskell.Prim
 
-record Thunk {ℓ} (a : @0 Size → Set ℓ) (@0 i : Size) : Set ℓ where
+record Thunk {ℓ} (a : Size → Set ℓ) (i : Size) : Set ℓ where
   constructor delay
   coinductive
   field force : {j : Size< i} → a j
--- a/lib/Haskell/Prim/Thunk.agda
+++ b/lib/Haskell/Prim/Thunk.agda
@@ -6,10 +6,10 @@ open import Agda.Builtin.Size public
 
 open import Haskell.Prim
 
-record Thunk {ℓ} (a : Size → Set ℓ) (i : Size) : Set ℓ where
+record Thunk {ℓ} (a : {j : Size} → Set ℓ) (i : Size) : Set ℓ where
   constructor delay
   coinductive
-  field force : {j : Size< i} → a j
+  field force : {j : Size< i} → a {j}
 open Thunk public
 
 {-# COMPILE AGDA2HS Thunk unboxed #-}
--- a/src/Agda2Hs/Compile/Term.hs
+++ b/src/Agda2Hs/Compile/Term.hs
@@ -195,11 +195,9 @@ compileTerm v = do
       True  -> compileErasedApp es
       False -> (`app` es) . Hs.Con () =<< hsQName (conName h)
     Lit l -> compileLiteral l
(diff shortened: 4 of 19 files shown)


jespercockx:
Making the `Size` arguments implicit is not ideal as Agda tends to eta-expand functions of type `{i : Size} -> Set` and then insert infinity instead of the size argument itself, effectively turning `T` into `\{i} -> T {infty}`. But this is just part of the general problem of size inference in Agda, so I think it's best to keep hoping this will eventually be fixed on the Agda side. Alternatively, we might want to look into supporting the old musical coinduction instead or in addition to sizes (but then I do not think it works well together with `--erased-cubical`).

lib/Haskell/Prim/Thunk.agda
```diff
@@ -6,10 +6,10 @@ open import Agda.Builtin.Size public
 
 open import Haskell.Prim
 
-record Thunk {ℓ} (a : @0 Size → Set ℓ) (@0 i : Size) : Set ℓ where
+record Thunk {ℓ} (a : {j : Size} → Set ℓ) (i : Size) : Set ℓ where
```
omelkonian, line 9: No need to bind this `j`, right?

lib/Haskell/Prim/Eq.agda
```diff
@@ -60,9 +60,13 @@ instance
...
   iEqList : ⦃ Eq a ⦄ → Eq (List a)
-  iEqList ._==_ []       []       = True
-  iEqList ._==_ (x ∷ xs) (y ∷ ys) = x == y && xs == ys
-  iEqList ._==_ _        _        = False
+  iEqList {a} ._==_ = eqList
+    where
+      eqList : List a → List a → Bool
+      eqList [] [] = True
+      eqList (x ∷ xs) (y ∷ ys) = x == y && eqList xs ys
+      eqList _ _ = False
```
omelkonian, line 68: So this is refactored to make the `--without-K` termination checker happy, but can we improve things in Agda itself or is there some fundamental limitation that does not allow for the previous more liberal definitions?

agda2hs.agda-lib
```diff
@@ -1,4 +1,4 @@
 name: agda2hs
 depend:
 include: lib
-flags: --erase-record-parameters
```
omelkonian, line 4: Why is this not needed anymore? I thought we want erased parameters to be actually erased.

Sure, for Sets/Levels/Sizes we have no annotations anymore, but what about everything else?

Or is it simply that just our Prim/Prelude modules don't require them, and it's up to the user to put this flag in case they use erased record parameters?

test/AllTests.agda
```diff
@@ -33,6 +33,7 @@ import LiteralPatterns
 import Issue92
 import HeightMirror
 import TransparentFun
+-- import StreamFusion TODO: cannot import b/c of --erased-cubical flag
```
omelkonian, line 36: Can you add some references to "size inference issues" in Agda, either in the test file or simply as a comment to the PR?

src/Agda2Hs/Compile/Term.hs
```diff
@@ -195,11 +195,9 @@ compileTerm v = do
       True  -> compileErasedApp es
       False -> (`app` es) . Hs.Con () =<< hsQName (conName h)
     Lit l -> compileLiteral l
-    Lam v b | usableModality v, getOrigin v == UserWritten -> do
-      unless (visible v) $ genericDocError =<< do
-        text "Implicit lambda not supported: " <+> prettyTCM (absName b)
+    Lam v b | keepArg v, getOrigin v == UserWritten -> do
       hsLambda (absName b) <$> underAbstr_ b compileTerm
```
omelkonian, line 199: Previously we gave an error on a hidden non-erased argument, but with this change we silently drop it. Is this intended?

src/Agda2Hs/Compile/Term.hs
```diff
@@ -195,11 +195,9 @@ compileTerm v = do
...
-    Lam v b | usableModality v ->
+    Lam v b | keepArg v ->
       -- System-inserted lambda, no need to preserve the name.
       underAbstraction_ b $ \ body -> do
         unless (visible v) $ genericDocError =<< do
```
omelkonian, line 203: Here, the error is kept (line 203), but it's actually dead code.

omelkonian (changes requested):

lib/Haskell/Prim/Thunk.agda
jespercockx: It's nice to give it a name so you can pass the argument in a named way (e.g. write `a {j = j}` instead of `a {j}`).

agda2hs.agda-lib
jespercockx: Yeah exactly, the primitive modules no longer require the flag, and the flag actually caused problems when I was trying to use Agda2Hs together with --erased-cubical.

test/AllTests.agda
jespercockx: The reason why I cannot import it here is not because of the bad inference of sizes, but simply because it uses the `--erased-cubical` flag while `AllTests` doesn't, but adding `--erased-cubical` to `AllTests` is not possible either because it also imports files that use `--erase-record-parameters`.

jespercockx: The proper thing to do here would be to have a separate file with tests using `--erased-cubical`

src/Agda2Hs/Compile/Term.hs
jespercockx: Yeah this is intended and is actually already the behaviour for implicit *arguments*. The typing rules enforce that the only implicit arguments and implicit lambdas that we can write are of type `Level`/`Set`/`Size`.

src/Agda2Hs/Compile/Term.hs
jespercockx: Right, I'll remove it

omelkonian (approved):
LGTM

PR #128, opened 2022-11-14
[ fix #120 ] Automatic language extension pragmas.

### LINKED ISSUE #120: Generate GHC language pragmas
Currently agda2hs sometimes generates code that requires -XScopedTypeVariables or -XBangPatterns. It would be good if it would automatically generate the appropriate language pragmas on top of the file when they are needed, so the user does not need to add them manually.

omelkonian:


jespercockx (approved):

omelkonian:
Just an observation on `compileTopLevelType`: it seems to be redoing work for typeclasses/instances in this code block:
```hs
    ...
    go (ExtendTel a atel) cont
      | isInstance a = do
          c <- Hs.TypeA () <$> compileType (unEl $ unDom a)
          underAbstraction a atel $ \tel ->
             go tel (cont . constrainType c)
      | otherwise = underAbstraction a atel $ \tel ->
          go tel (cont . qualifyType (absName atel))
```
Everything works, but I believe there is some code duplication or even repeated computation.

jespercockx:
I'm not sure what is the code duplication that you are talking about. However, if you see a way to simplify the code feel free to make another PR.

PR #129, opened 2022-11-15
Fix Issue 127

### LINKED ISSUE #127: Reject datatype parameter of kind other than *
For example:
```agda
data T (n : Nat) : Set where
  MkT : T n
{-# COMPILE AGDA2HS T #-}
```
compiles to
```
data T n = MkT
```

Riib11:
Fixes #127

Previously, the user could include any type index among the telescope of arguments in a record or datatype definition, and agda2hs would compile it as if it had type star even if it had a type that should not be compiled to star, such as `Nat`.

This PR fixes this issue by modifying `Agda2Hs.Compile.Type.compileTeleBinds` to additionally check that the type of an argument in the telescope is of the form `Sort (Type _)`. Otherwise, it emits a new error message: `Kind of bound argument not supported: (<arg name> : <arg type>)`.

Two fail tests have been added which test a simple datatype definition and record:
- `tests/Fail/NonStarDatatypeIndex.agda`
- `tests/Fail/NonStarRecordIndex.agda`


.gitignore
```diff
@@ -4,3 +4,6 @@ dist-newstyle
 test/build
 test/agda2hs
 *~
+docs/build/
+test/*.hs
+test/Haskell
```
jespercockx, line 9: please make sure there's a newline at the end of the file

test/Test.hs
```diff
@@ -0,0 +1,152 @@
+{-# LANGUAGE LambdaCase #-}
```
jespercockx, line 1: This file shouldn't be included in the repo, it already lives in test/golden

src/Agda2Hs/Compile/Type.hs
```diff
@@ -161,6 +162,16 @@ compileDom x a
...
+    $ teleArgNames tel `zip` (flattenTel tel :: [Dom Type]) )
+    $ \(x, t) -> do
+      checkValidTyVarName x
+      case unEl t of
+        Sort (Type _) -> return $ Hs.UnkindedVar () x
+        _ -> genericDocError =<< do
+          text "Kind of bound argument not supported:" <+> 
+            parens (text (Hs.prettyPrint x) <> text " : " <> prettyTCM t)
+
+
```
jespercockx, line 177: Consider using https://github.com/agda/fix-whitespace to make sure your whitespace is uniform

src/Agda2Hs/Compile/Type.hs
```diff
@@ -161,6 +162,16 @@ compileDom x a
 
 compileTeleBinds :: Telescope -> C [Hs.TyVarBind ()]
 compileTeleBinds tel =
-  forM (map (hsName . unArg) $ filter keepArg $ teleArgNames tel) $ \x -> do
-    checkValidTyVar x
-    return $ Hs.UnkindedVar () x
+  forM
+    ( fmap (\(argName, domType) -> (hsName . unArg $ argName, unDom domType))
```
jespercockx, line 166: Could you factor this out into a (local) function to make it more readable?

jespercockx: Also, the `fmap` and `filter`could be combined in a call to `mapMaybe`

src/Agda2Hs/Compile/Type.hs
```diff
@@ -161,6 +162,16 @@ compileDom x a
...
+    . filter (\(argName, _domType) -> keepArg argName) 
+    $ teleArgNames tel `zip` (flattenTel tel :: [Dom Type]) )
+    $ \(x, t) -> do
+      checkValidTyVarName x
+      case unEl t of
+        Sort (Type _) -> return $ Hs.UnkindedVar () x
+        _ -> genericDocError =<< do
+          text "Kind of bound argument not supported:" <+> 
+            parens (text (Hs.prettyPrint x) <> text " : " <> prettyTCM t)
```
jespercockx, line 175: Could we factor out the check on the sort into a standalone top-level function? It seems likely that we'll want to extend it with other things in the future.

jespercockx (changes requested):

jespercockx:
It seems the test output in `test/golden` has changed, could you run `make golden` and push the changes?

jespercockx:
I'm sorry for letting this hang for so long! I've just rebased this on master and resolved the conflicts, if CI passes then this can be merged.

jespercockx (approved):

PR #130, opened 2022-11-15
Pattern matching with Nat literals

uncle-betty:
A proposal to add pattern matching with Nat literals to `agda2hs`.

Includes the factorial and Fibonacci function over `Nat` as examples.

Just an idea. Feel free to close, if this is too specific to my use cases.

Coming up with the recursion schemes for the examples included quite a bit of trial and error. And I still don't fully understand, why Agda accepts, for example, the `>1` constructor in the last clause of `fib`. Agda seems to magically rule out `≡0` and `≡1`, so that `>1` is the only option and thus doesn't complain about a pattern match with an erased function argument. Same for the `≢0` in the last clause of `fac`.

In any case, this seems to address my need for basic "count down from n to 0" Nat recursion.

### DIFF
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -46,6 +46,8 @@ isSpecialPat = prettyShow >>> \ case
   "Haskell.Prim.Tuple._;_" -> Just tuplePat
   "Agda.Builtin.Int.Int.pos" -> Just posIntPat
   "Agda.Builtin.Int.Int.negsuc" -> Just negSucIntPat
+  "Agda.Builtin.Nat.Nat.zero" -> Just zeroNatPat
+  "Agda.Builtin.Nat.Nat.suc" -> Just sucNatPat
   _ -> Nothing
 
 isUnboxCopattern :: DeBruijnPattern -> C Bool
@@ -85,6 +87,15 @@ negSucIntPat c i [p] = do
   n <- (1+) <$> compileLitNatPat (namedArg p)
   return $ Hs.PLit () (Hs.Negative ()) (Hs.Int () n (show (negate n)))
 negSucIntPat _ _ _ = __IMPOSSIBLE__
+
+zeroNatPat :: ConHead -> ConPatternInfo -> [NamedArg DeBruijnPattern] -> C (Hs.Pat ())
+zeroNatPat _ _ _ = return $ Hs.PLit () (Hs.Signless ()) (Hs.Int () 0 "0")
+
+sucNatPat :: ConHead -> ConPatternInfo -> [NamedArg DeBruijnPattern] -> C (Hs.Pat ())
+sucNatPat _ _ [p] = do
+  n <- (1+) <$> compileLitNatPat (namedArg p)
+  return $ Hs.PLit () (Hs.Signless ()) (Hs.Int () n (show n))
+sucNatPat _ _ _ = __IMPOSSIBLE__
 
 -- The bool argument says whether we also want the type signature or just the body
 compileFun, compileFun' :: Bool -> Definition -> C [Hs.Decl ()]
--- a/test/AllTests.agda
+++ b/test/AllTests.agda
@@ -39,6 +39,7 @@ import Issue94
 import Issue107
 import Importer
 import DoNotation
+import NatLitPat
 
 {-# FOREIGN AGDA2HS
 import Issue14
@@ -78,4 +79,5 @@ import BangPatterns
 import Issue94
 import Importer
 import DoNotation
+import NatLitPat
 #-}
--- a/test/NatLitPat.agda
+++ b/test/NatLitPat.agda
@@ -1,0 +1,43 @@
+module NatLitPat where
+
+open import Agda.Builtin.Equality using (_≡_ ; refl)
+open import Haskell.Prelude
+
+data RecHelp₁ : Nat → Set where
+  ≡0 : RecHelp₁ 0
+  ≢0 : {n : Nat} → ⦃ _ : Num.MinusOK iNumNat n 1 ⦄ → RecHelp₁ (n - 1) → RecHelp₁ n
+
+data RecHelp₂ : Nat → Set where
+  ≡0 : RecHelp₂ 0
+  ≡1 : RecHelp₂ 1
+  >1 : {n : Nat} → ⦃ _ : Num.MinusOK iNumNat n 1 ⦄ → ⦃ _ : Num.MinusOK iNumNat n 2 ⦄ →
+       RecHelp₂ (n - 1) → RecHelp₂ (n - 2) → RecHelp₂ n
+
+instance
+  @0 recHelp₁ : {n : Nat} → RecHelp₁ n
+  recHelp₁ {zero}  = ≡0
+  recHelp₁ {suc n} = ≢0 (recHelp₁ {n})
+
+  @0 recHelp₂ : {n : Nat} → RecHelp₂ n
+  recHelp₂ {zero}        = ≡0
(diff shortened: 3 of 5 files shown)


jespercockx:
Thank you for the PR! I'm happy to accept this, but it seems your approach to write recursive functions would work equally well for `Int` as for `Nat`, since all of the actual recursion is happening in the (erased) proof. Still, being able to pattern match on `Natural` literals seems like a strict improvement over the status quo!

jespercockx:
Actually, is the special case for `sucNatPat` really needed? Since the only case that it actually handles is `suc` applied to a literal, it seems it would already be covered by `compileLitNatPat`. Could you try to remove it and see if your test case still goes through?

uncle-betty:
Oooooh, that's funny! I had become so obsessed with `Nat`s that I completely missed that I now had something on my hands that would also work with `Int`s.

Thank you for the review. I tried removing the `suc` case from `isSpecialPat` and the test fails.

As far as I can tell, `compileLitNatPat` is otherwise only used for `Nat`s wrapped in `pos` and `negsuc`, respectively - not for naked `Nat`s.

When the `semantics` function for each special pattern is applied, the outermost constructor has already been stripped. That's why I needed a `semantics` function for `suc` that would make up for the missing outermost `suc` by adding 1 to the resulting value.

uncle-betty:
Ah, wait! I'm not so sure any longer that the recursion scheme would actually work with `Integer`s. I tried this, which lumps in negative integers with zero:

```
data RecHelp₃ : Integer → Set where
  ≡0 : {i : Integer} → RecHelp₃ i
  ≢0 : {i : Integer} → RecHelp₃ (i - 1) → RecHelp₃ i

open Integer

instance
  @0 recHelp₃ : {i : Integer} → RecHelp₃ i
  recHelp₃ {pos zero}    = ≡0
  recHelp₃ {pos (suc n)} = ≢0 (recHelp₃ {pos n})
  recHelp₃ {negsuc n}    = ≡0

fac' : (i : Integer) → ⦃ @0 _ : RecHelp₃ i ⦄ → Integer
fac' (pos 0)          = 1
fac' n       ⦃ ≢0 r ⦄ = n * fac' (n - 1)
```

And now I do get `Cannot branch on erased argument ...` for the second clause of `fac'`.

I guess it's about now having two constructors that produce a `RecHelp₃ i` for arbitrary `i`. In the `Nat` case, there's only ever one of these constructors. The other ones apply to specific `Nat`s - 0 or 1.

jespercockx:
Here is how I would implement your test case with `Integer` instead of `Nat`:
```agda
open import Agda.Builtin.Equality using (_≡_ ; refl)
open import Haskell.Prelude

variable
  n : Nat
  i : Integer

data PosInteger : Integer → Set where
  pos : (n : Nat) → PosInteger (Integer.pos n)

instance
  _ : PosInteger (Integer.pos n)
  _ = pos _

data RecHelp₁ : Integer → Set where
  ≡0 : RecHelp₁ 0
  >0 : RecHelp₁ (Integer.pos n) → RecHelp₁ (Integer.pos (suc n))

data RecHelp₂ : Integer → Set where
  ≡0 : RecHelp₂ 0
  ≡1 : RecHelp₂ 1
  >1 : RecHelp₂ (i - 1) → RecHelp₂ (i - 2) → RecHelp₂ i

instance
  @0 recHelp₁ : {{PosInteger i}} → RecHelp₁ i
  recHelp₁ {{pos zero}}    = ≡0
  recHelp₁ {{pos (suc n)}} = >0 (recHelp₁ {{pos n}})

  @0 recHelp₂ : {{PosInteger i}} → RecHelp₂ i
  recHelp₂ {{pos zero}}          = ≡0
  recHelp₂ {{pos (suc zero)}}    = ≡1
  recHelp₂ {{pos (suc (suc n))}} = >1 (recHelp₂ {{pos (suc n)}}) (recHelp₂ {{pos n}})

fac : (i : Integer) → @0 ⦃ RecHelp₁ i ⦄ → Integer
fac (Integer.pos 0) = 1
fac i {{>0 r}} = i * fac (i - 1) ⦃ r ⦄

{-# COMPILE AGDA2HS fac #-}

fib : (i : Integer) → ⦃ @0 _ : RecHelp₂ i ⦄ → Integer
fib (Integer.pos 0) = 0
fib (Integer.pos 1) = 1
fib n {{>1 r₁ r₂}}  = fib (n - 1) ⦃ r₁ ⦄ + fib (n - 2) ⦃ r₂ ⦄

{-# COMPILE AGDA2HS fib #-}

proof₁ : fac 5 ≡ 120
proof₁ = refl

proof₂ : fib 10 ≡ 55
proof₂ = refl
```
However, this currently does not seem to be accepted by agda2hs either, I get the following error:
```
not a literal natural number pattern: n
```
I'll take a look at what's going wrong.

uncle-betty:
Oooooh, clever! It hadn't occurred to me to restrict the kind of integers that could be passed to `recHelp` to `pos`. And it dovetails nicely with the overall idea to make `recHelp` do the heavy lifting so that `fac` and `fib` remain translatable by `agda2hs`.

Yes, while `agda2hs` only supports `Nat` literals for `pos` and `negsuc`, it does seem strange that your solution would hit that limitation. Odd.

jespercockx:
So the problem seems to be that internally Agda has elaborated the pattern `fac i {{>0 r}}` to `fac (Integer.pos .(suc n)) {{>0 {n} r}}`. It took me a while, but I managed to come up with a version that does not have this problem:
```agda

open import Agda.Builtin.Equality using (_≡_ ; refl)
open import Haskell.Prelude

variable
  @0 n : Nat
  @0 i : Integer

data PosInteger : Integer → Set where
  zero : PosInteger 0
  suc  : PosInteger (i - 1) → PosInteger i

instance
  @0 posIsPos : PosInteger (Integer.pos n)
  posIsPos {zero}  = zero
  posIsPos {suc n} = suc (posIsPos {n})

fac : (i : Integer) → @0 ⦃ PosInteger i ⦄ → Integer
fac (Integer.pos 0) = 1
fac i {{suc r}}     = i * fac (i - 1) ⦃ r ⦄

{-# COMPILE AGDA2HS fac #-}

proof₁ : fac 5 ≡ 120
proof₁ = refl

data FibHelper : Integer → Set where
  zero : FibHelper 0
  one  : FibHelper 1
  more : FibHelper (i - 2) → FibHelper (i - 1) → FibHelper i

instance
  @0 fibHelper : FibHelper (Integer.pos n)
  fibHelper {zero} = zero
  fibHelper {suc zero} = one
  fibHelper {suc (suc n)} = more (fibHelper {n}) (fibHelper {suc n})

fib : (i : Integer) → @0 ⦃ FibHelper i ⦄ → Integer
fib (Integer.pos 0) = 0
fib (Integer.pos 1) = 1
fib i {{more r₁ r₂}}  = fib (i - 2) ⦃ r₁ ⦄ + fib (i - 1) ⦃ r₂ ⦄

{-# COMPILE AGDA2HS fib #-}

proof₂ : fib 10 ≡ 55
proof₂ = refl
```
Since is kind of non-trivial to come up with on the spot, it would be good to document this somewhere. (Pinging @Riib11 who is currently working on the documentation of agda2hs).

uncle-betty:
"kind of non-trivial"  - understatement of the month! :)

Then, maybe, instead of merging this PR, simply add your version to the `test` directory? That's where I usually go looking for help when I'm stuck.

Also, would the basic "recurse from n down to 0" scheme be common enough to be useful to more people? If so, then would it make sense to add renamed versions of `FacHelper` and `facHelper` to the `agda2hs` prelude? Then people wouldn't have to come up with their own.

PR #72, opened 2021-06-14
[new] support default class methods

omelkonian:

PR #136, opened 2023-02-16
Add support for qualified imports

### LINKED ISSUE #123: Qualified imports
Currently all imports that are generated by agda2hs are *unqualified*, even if the Agda source uses a qualified import. We should instead respect the user decision on whether to use a qualified or unqualified name.

Note that the current approach is actually buggy when two modules are imported qualified that export the same name.

jespercockx:
This fixes #123 by trying to detect which names are used in a qualified way through the reverse scope map that's used internally by Agda. There are a few dirty tricks here to ensure that names of internal Agda modules (e.g. record modules) do not appear in the generated Haskell code, but I *think* my approach does the right thing in all "normal" use cases.

One thing that could still be improved here is that agda2hs could throw an error when you use local modules in your Agda code and you do not `open public` them. Currently this will just lead to agda2hs generating bogus qualifiers. For this reason I also had to move the functions that are defined in the `Foldable` class out of the record, a proper solution here would be to instead make use of the support for default methods added in https://github.com/agda/agda2hs/pull/72.

The main thing this still needs is more testing, so if you can please write some code with qualified imports and report back on if it works!

### DIFF
--- a/lib/Haskell/Prim/Foldable.agda
+++ b/lib/Haskell/Prim/Foldable.agda
@@ -19,54 +19,54 @@ record Foldable (t : Set → Set) : Set₁ where
   field
     foldMap : ⦃ Monoid b ⦄ → (a → b) → t a → b
 
-  foldr : (a → b → b) → b → t a → b
-  foldr f z t = foldMap ⦃ MonoidEndo ⦄ f t z
-
-  foldl : (b → a → b) → b → t a → b
-  foldl f z t = foldMap ⦃ MonoidEndoᵒᵖ ⦄ (flip f) t z
-
-  any : (a → Bool) → t a → Bool
-  any = foldMap ⦃ MonoidDisj ⦄
-
-  all : (a → Bool) → t a → Bool
-  all = foldMap ⦃ MonoidConj ⦄
-
-  and : t Bool → Bool
-  and = all id
-
-  or : t Bool → Bool
-  or = any id
-
-  null : t a → Bool
-  null = all (const False)
-
-  concat : t (List a) → List a
-  concat = foldMap id
-
-  concatMap : (a → List b) → t a → List b
-  concatMap = foldMap
-
-  elem : ⦃ Eq a ⦄ → a → t a → Bool
-  elem x = foldMap ⦃ MonoidDisj ⦄ (x ==_)
-
-  notElem : ⦃ Eq a ⦄ → a → t a → Bool
-  notElem x t = not (elem x t)
-
-  toList : t a → List a
-  toList = foldr _∷_ []
-
-  sum : ⦃ iNum : Num a ⦄ → t a → a
-  sum = foldMap ⦃ MonoidSum ⦄ id
-
-  product : ⦃ iNum : Num a ⦄ → t a → a
-  product = foldMap ⦃ MonoidProduct ⦄ id
-
-  length : t a → Int
-  length = foldMap ⦃ MonoidSum ⦄ (const 1)
-
 open Foldable ⦃ ... ⦄ public
 
 {-# COMPILE AGDA2HS Foldable existing-class #-}
+
+foldr : ⦃ Foldable t ⦄ → (a → b → b) → b → t a → b
+foldr f z t = foldMap ⦃ it ⦄ ⦃ MonoidEndo ⦄ f t z
+
+foldl : ⦃ Foldable t ⦄ → (b → a → b) → b → t a → b
+foldl f z t = foldMap ⦃ it ⦄ ⦃ MonoidEndoᵒᵖ ⦄ (flip f) t z
+
+any : ⦃ Foldable t ⦄ → (a → Bool) → t a → Bool
+any = foldMap ⦃ it ⦄ ⦃ MonoidDisj ⦄
+
+all : ⦃ Foldable t ⦄ → (a → Bool) → t a → Bool
+all = foldMap ⦃ it ⦄ ⦃ MonoidConj ⦄
+
+and : ⦃ Foldable t ⦄ → t Bool → Bool
+and = all id
+
+or : ⦃ Foldable t ⦄ → t Bool → Bool
+or = any id
+
+null : ⦃ Foldable t ⦄ → t a → Bool
+null = all (const False)
+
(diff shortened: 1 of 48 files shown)


src/Agda2Hs/Compile/Utils.hs
```diff
@@ -142,8 +145,12 @@ isClassFunction q
           ExistingClassPragma -> True
           _                   -> False
       _                             -> return False
-  where
-    m = qnameModule q
+
+-- Drops the last (record) module for typeclass methods
+dropClassModule :: ModuleName -> C ModuleName
```
omelkonian, line 150: This is not used anywhere, did you have something in mind for the future?

lib/Haskell/Prim/Foldable.agda
```diff
@@ -19,54 +19,54 @@ record Foldable (t : Set → Set) : Set₁ where
...
+open Foldable ⦃ ... ⦄ public
 
-  foldl : (b → a → b) → b → t a → b
-  foldl f z t = foldMap ⦃ MonoidEndoᵒᵖ ⦄ (flip f) t z
+{-# COMPILE AGDA2HS Foldable existing-class #-}
 
-  any : (a → Bool) → t a → Bool
-  any = foldMap ⦃ MonoidDisj ⦄
+foldr : ⦃ Foldable t ⦄ → (a → b → b) → b → t a → b
+foldr f z t = foldMap ⦃ it ⦄ ⦃ MonoidEndo ⦄ f t z
```
omelkonian, line 27: I think this has to be resolved before making it to the release. You mention the way to go is 'default methods', but we only care about this when we have multiple minimal records, do you mean that we should apply the same treatment here to properly resolve this and put the definitions back into the record?

omelkonian (commented):
Cleaned up the code a bit, will write some more comprehensive tests tomorrow.

src/Agda2Hs/Compile/Name.hs
```diff
@@ -82,7 +98,26 @@ compileQName f
         -> return $ Just $ unArg rt
       _ -> return Nothing
 
-compileModuleName :: Monad m => ModuleName -> m (Hs.ModuleName ())
+    isQualified :: QName -> Hs.ModuleName () -> C IsQualified
+    isQualified f mod = (inverseScopeLookupName f <$> getScope) >>= \case
```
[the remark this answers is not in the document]
jespercockx: Indeed, it was just hiding under a rock.

src/Agda2Hs/Compile/Utils.hs
jespercockx: Not really, I think I was using it at some point but then I refactored something and it was no longer needed.

lib/Haskell/Prim/Foldable.agda
jespercockx: Yeah I was thinking of applying the same method here. Even if there is only a single minimal record, having a version of the record where all of the methods are fields still makes sense since this allows the defaults to be overridden.

jespercockx:
@omelkonian Thanks for the cleaning up of the code!

src/Agda2Hs/Compile/Utils.hs
jespercockx: Actually, perhaps we could use it to make sure we do not add qualifiers for record modules.

omelkonian: I managed to work around these issues without this function, so I deleted it in the latest push.

lib/Haskell/Prim/Foldable.agda
omelkonian: I've refactored `lib/` to use proper default methods, it makes things a bit clunky to use (because you now need to use the record-as-module syntax to construct minimal instances) but it works. It would be nice to someday make the instance translation a bit more liberal to allow for more complicated expression that may streamline such default constructions (e.g. see `foldMap=_` in my commit).

src/Agda2Hs/Compile/Name.hs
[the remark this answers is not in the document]
omelkonian: This still doesn't cover the case for `hiding`, which leads us back to the recurrent problem of associating the abstract syntax with the original concrete syntax to retrieve such information. 

I think we can merge this feature as only supporting qualifications, but I want to just keep a note here in case we do find a more complete overall  solution to this problem (e.g. by tweaking Agda to keep such information for needy backends).

omelkonian:
@jespercockx I have updated the code and I think it is now sufficiently complete, albeit still brittle. Can you review my code and merge if everything seems fine?

src/Agda2Hs/Compile/Name.hs
```diff
@@ -30,65 +32,81 @@ import Agda2Hs.Compile.Types
 import Agda2Hs.Compile.Utils
 import Agda2Hs.HsUtils
 
-isSpecialName :: QName -> Maybe (Hs.QName (), Maybe Import)
+isSpecialCon :: QName -> Maybe (Hs.QName ())
+isSpecialCon = prettyShow >>> \case
+    "Agda.Builtin.List.List"     -> special Hs.ListCon
```
jespercockx, line 37: Why is List here while other type constructors are in isSpecialName?

jespercockx (commented):
Should the automatic -XLambdaCase be part of this PR?

src/Agda2Hs/Compile/Name.hs
omelkonian: This separately deals with all "special" constructors in syntax (c.f. the [haskell-src-ext docs](https://hackage.haskell.org/package/haskell-src-exts-1.23.1/docs/Language-Haskell-Exts-Syntax.html#t:SpecialCon)]).

omelkonian:
> Should the automatic -XLambdaCase be part of this PR?

Yeah, not really. Let me spin off a separate PR for that.

omelkonian:
@jespercockx Any idea how to get the proper error location for `test/Fail/ClashingImports.agda`?

jespercockx:
> @jespercockx Any idea how to get the proper error location for test/Fail/ClashingImports.agda?

I am not sure I understand what exactly the problem is here. I'd like to merge this now, so could you please create a new issue about the bad error message?

PR #144, opened 2023-02-22
[ fix #143 ] allow function kind

### LINKED ISSUE #143: Allow higher-kinded records/datatypes
Just tried defining my own record (compiled to a Haskell datatype) parametrized by some monad, and was dumbfounded that it is somehow rejected, as parameters are apparently only allowed to have kind `*` (i.e `Set`).

```agda
record Kleisli (m : Set → Set) (a b : Set) : Set where
  constructor MkKleisli ; field runKleisli : a → m b
{-# COMPILE AGDA2HS Kleisli #-}
```

This throws the following error: `Kind not supported: Set → Set`, from [there](https://github.com/agda/agda2hs/blob/master/src/Agda2Hs/Compile/Type.hs#L199).

flupe:
Added support for kind of type families. The error reporting felt somewhat redundant so I removed part of it. ~~For now kinds that are not `*` are annotated in the output Haskell, but maybe we never want annotations?~~

Next:
- [x] Add tests.
- [ ] I suspect more checks have to be done when compiling the Type to a Haskell kind, like making sure that the domain doesn't have some weird modality? Wouldn't mind some comments on this.

### DIFF
--- a/src/Agda2Hs/Compile/Type.hs
+++ b/src/Agda2Hs/Compile/Type.hs
@@ -183,8 +183,7 @@ compileKeptTeleBind :: Hs.Name () -> Type -> C (Hs.TyVarBind ())
 compileKeptTeleBind x t = do
   checkValidTyVarName x
   case compileKind t of
-    Just (Hs.TyStar ()) -> pure $ Hs.UnkindedVar () x
-    Just k              -> pure $ Hs.KindedVar () x k
+    Just k              -> pure $ Hs.UnkindedVar () x -- In the future we may want to show kind annotations
     _                   -> genericDocError =<<
       text "Kind of bound argument not supported:"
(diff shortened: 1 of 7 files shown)


jespercockx:
I guess this is implementing https://github.com/agda/agda2hs/issues/38?

jespercockx:
Since types and kinds are the same thing in (modern) Haskell, perhaps it would make sense to also merge the two functions `compileType` and `compileKind`. Or at least, I don't immediately see what would go wrong if we did that.

jespercockx:
@flupe Could you bring this back up to date with current master? Then we can merge it.

flupe:
Rebased.

> Since types and kinds are the same thing in (modern) Haskell, perhaps it would make sense to also merge the two functions compileType and compileKind.

Indeed, I'll let you open an issue for that.

PR #148, opened 2023-02-23
[ fix #147 ] preserve do notation when pattern matching on bind

### LINKED ISSUE #147: Do notation and pattern-matching bind
#101 added support for do-notation, but it only works if the second argument to the monadic bind is a lambda. If one pattern-matches on a monadic bind using do-notation, it gets converted into a `\where`, so we loose the do notation.

```agda
do
  (x , y) <- smthg
  pure x
```

Becomes:

```agda
smthg >>= \case
  (x , y) -> pure x
```

flupe:


src/Agda2Hs/Compile/Term.hs
```diff
@@ -123,12 +123,16 @@ bind :: QName -> Elims -> C (Hs.Exp ())
 bind q (e:es) = do
   checkInstance $ unArg $ isApplyElim' __IMPOSSIBLE__ e
   compileElims es >>= \case
-    (u : Hs.Lambda _ [p] v : vs) -> do
-      let stmt1 = Hs.Generator () p u
-      case v of
-        Hs.Do _ stmts -> return $ Hs.Do () (stmt1 : stmts)
-        _             -> return $ Hs.Do () [stmt1, Hs.Qualifier () v]
+    [u, Hs.Lambda _ [p] v] -> pure (bind' u p v)
+    [u, Hs.LCase () [Hs.Alt () p (Hs.UnGuardedRhs () v) Nothing]] -> pure (bind' u p v)
```
omelkonian, line 127: Previously the case of lengths >2 were also covered (although `vs` was just ignored). Have you considered this and fixed it or are we missing any corner case?

flupe: I think having `vs /= []` is impossible to reach given how these (extended) lambdas are generated by Agda. It seemed more principled to explicitly expect exactly 2 arguments for the `do` notation, and fall back on the last case with desugared `>>=` if there is more than 2 arguments, but alas I think it doesn't matter.

PR #149, opened 2023-02-23
Automatic -XLambdaCase

omelkonian:


flupe:
Pattern-matching lambdas also get generated and possibly removed when compiling and processing the `do` notation. I guess you could also add a `compilingDo` in the context like the rest, but maybe another alternative would be to carry a state `Map KnownExtension Int` or something, and increase/decrease count when LambdaCase get generated/removed. In the end, only introduce extensions with non-zero count? Maybe it's overkill, but may be a bit more robust and future-proof.

omelkonian:
> I guess you could also add a `compilingDo` in the context like the rest, but maybe another alternative would be to carry a state `Map KnownExtension Int` or something, and increase/decrease count when LambdaCase get generated/removed.

I've generalised the compiling flags to make this a bit less boilerplate-y, but I agree your suggestion is more robust because every time you see a `Hs.LCase` in the code you know you have to either add or subtrack 1 from the counter, whether in my current solution it's more spaghetti.

src/Agda2Hs/Compile/Term.hs
```diff
@@ -180,6 +179,8 @@ lambdaCase q es = setCurrentRange (nameBindingSite $ qnameName q) $ do
     -- simply return the body.
     [Hs.Match _ _ [] (Hs.UnGuardedRhs _ rhs) _] -> return rhs
     _ -> do
+      unlessM (fromMaybe False . fmap (`elem` [CaseOf, MonadicBind]) <$> asks isCompiling) $
```
flupe, line 182: Could you use `isCompilingCaseOrBind` to make it more legible?

flupe (approved):
Looks good!

src/Agda2Hs/Compile/Term.hs
omelkonian: Oops, leftover.

omelkonian:
@flupe Now I am wondering whether there is a corner case that breaks this, something like a λ-case deeper inside a monadic bind or case expression, which would prevent the insertion of -XLambdaCase incorrectly.

Let me check it and make sure this does not happen, otherwise fall back to your `State`-ful suggestion.

omelkonian:
Indeed I broke it:
```agda
mhead2 : List a → (Maybe a → Maybe a)
mhead2 xs = case xs of λ where
  []      → λ where Nothing → Nothing; (Just _) → Nothing
  (x ∷ _) → λ where Nothing → Nothing; (Just _) → Just x
{-# COMPILE AGDA2HS mhead2 #-}
```

Changing to your solution.

omelkonian:
@flupe Lemme revert the changes with `compiling`, no need for it anymore.

omelkonian:
Rebased on current HEAD.

src/Agda2Hs/Compile/Term.hs
```diff
@@ -180,6 +181,7 @@ lambdaCase q es = setCurrentRange (nameBindingSite $ qnameName q) $ do
     -- simply return the body.
     [Hs.Match _ _ [] (Hs.UnGuardedRhs _ rhs) _] -> return rhs
     _ -> do
+      incrementLCase
       alts <- mapM clauseToAlt cs -- Pattern lambdas cannot have where blocks
       args <- compileElims rest
       return $ eApp (Hs.LCase () alts) args
```
flupe, line 187: Could you use the `hsLCase` defined on line 214 of `Compile.Types` so that its comment holds?

PR #151, opened 2023-02-24
[ fix #146 ] prevent copatterns in instance field clause

### LINKED ISSUE #146: Copatterns not prevented when defining class instances
Since d71c9e61cc copatterns are explicitely forbidden in definitions, except for instance declarations. However, if an instance field uses copatterns, they are not prevented which leads to incorrect Haskell code, again.

```agda
record Wrap (a : Set) : Set where
  constructor MkWrap
  field wrapped : a
open Wrap public
{-# COMPILE AGDA2HS Wrap #-}

record Class (a : Set) : Set where
  field
    method : Wrap a → Wrap a
open Class ⦃...⦄ public
{-# COMPILE AGDA2HS Class class #-}

instance
  BoolClass : Class Bool
  BoolClass .method (MkWrap x) .wrapped = x
  {-# COMPILE AGDA2HS BoolClass #-}
```

```haskell
data Wrap a = MkWrap{wrapped :: a}

class Class a where
    method :: Wrap a -> Wrap a

instance Class Bool where
    method (MkWrap x) Testing.Wrap.wrapped = x
```

Side note: using copatterns for records with a single field should be *fine*, we could probably allow it, possibly only for newtypes #141 for starters.

flupe:
Very simple fix, but the naming could be improved upon.
Let's see how #149 evolves before improving this PR.

I suspect we could rename `isCompilingInstance` to `copatternsEnabled` and have switches `enableCopatterns`/`disableCopatterns` rather than `compilingInstance` and the badly named `outsideCompilationInstance`.

### DIFF
--- a/src/Agda2Hs/Compile/Type.hs
+++ b/src/Agda2Hs/Compile/Type.hs
@@ -3,7 +3,7 @@
 module Agda2Hs.Compile.Type where
 
 import Control.Arrow ( (>>>) )
-import Control.Monad ( forM )
+import Control.Monad ( forM, when )
 import Control.Monad.Reader ( asks )
 import Data.Maybe ( mapMaybe )
 
@@ -87,8 +87,8 @@ qualifyType s = \case
 -- (if any) as explicitly bound type arguments.
 -- The continuation is called in an extended context with these type
 -- arguments bound.
-compileTopLevelType :: Type -> (Hs.Type () -> C a) -> C a
-compileTopLevelType t cont = do
+compileTopLevelType :: Bool -> Type -> (Hs.Type () -> C a) -> C a
+compileTopLevelType keepType t cont = do
     ctxArgs <- getContextArgs
     modTel <- lookupSection =<< currentModule
     go (modTel `apply` ctxArgs) cont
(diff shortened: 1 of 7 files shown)


omelkonian:
Doesn't seem like #149 will affect `isCompilingInstance` after all, so go ahead and rename to `copatternsEnabled` which is nicer.

src/Agda2Hs/Compile/Type.hs
```diff
@@ -104,8 +104,7 @@ compileTopLevelType t cont = do
           underAbstraction a atel $ \tel ->
             go tel (cont . constrainType c)
       | otherwise = underAbstraction a atel $ \tel -> do
-          unlessM (asks isCompilingInstance) $
```
flupe, line 107: This was the only other place using `isCompilingInstance` not pertaining to copatterns. As it appears, it's not even related to instances, but rather to whether or not the generated Haskell type will be used in the output code.
When compiling a class instance, we discard the type and thus shouldn't enable the language extension. But that's not the only place where we discard the type (Every `compileFun False` does the same), so I added a flag. A bit ad-hoc though.

src/Agda2Hs/Compile/Type.hs
```diff
@@ -3,7 +3,7 @@
 module Agda2Hs.Compile.Type where
 
 import Control.Arrow ( (>>>) )
-import Control.Monad ( forM )
+import Control.Monad ( forM, when )
```
omelkonian, line 6: This is a leftover I gather.

PR #158, opened 2023-03-09
Add support for custom fixity declarations

flupe:
Implements #41.

WIP, I have yet to fix infix applications parentheses.

### DIFF
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -28,6 +28,7 @@ import Agda.TypeChecking.Sort ( ifIsSort )
 
 import Agda.Utils.Functor ( (<&>) )
 import Agda.Utils.Impossible ( __IMPOSSIBLE__ )
+import Agda.Utils.Lens ((^.))
 import Agda.Utils.List
 import Agda.Utils.Maybe
 import Agda.Utils.Monad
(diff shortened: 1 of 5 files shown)


flupe:
Kinda hitting a roadblock here. Currently pretty-printing infix applications (`1 + 2 - 5` instead of `1 + (2 - 5)`) is done by processing the generated Haskell syntax and accounting for the default fixities of things in the prelude (this mapping is exported by the Haskell syntax library we use).

My PR works but fixity of custom operators is ignored and each application sees its children parenthesized (which is fine).

In order to *also* pretty-print custom operators, I wanted the post-processing pass to rely on a mapping from names to fixity of things *in scope of the current module*. I thought this info could be found in `_scopeFixities` from [`ScopeInfo`](https://hackage.haskell.org/package/Agda-2.6.3/docs/Agda-Syntax-Scope-Base.html#t:ScopeInfo), info we retrieve from `iInsideScope` of the module [`Interface`](https://hackage.haskell.org/package/Agda-2.6.3/docs/Agda-Compiler-Backend.html#t:Interface) that we get when compiling the module. To my surprise, this mapping is always empty, which I find surprising.

Looking for some comments on this, @jespercockx ? The lazy way out is to not pretty-print properly custom operators, which is arguably fine is the point of Agda2HS is to make it easy to bind to generated Haskell, not make said Haskell particularly pretty.

jespercockx:
Hm that's a bit surprising indeed, I'm not sure why the fixity information is absent. I guess this will require some detective work in the Agda codebase itself to figure out where the fixity information is going to.

I put this PR on the 1.0 milestone because I assumed it was (almost) ready, but since it isn't I propose we keep the current situation as it is.

flupe:
Just rebased this PR on current master. Just wanted to note that it does work currently (you can introduce your own fixity). The only caveat is that your custom declarations won't be used when pretty-printing the generated Haskell code, but it will generate correct Haskell.

jespercockx:
Sounds good, thanks for the PR!

PR #161, opened 2023-03-23
Add standalone `deriving`

sarajuhosova:
This PR adds a way to derive type class instances which are available on the Agda side.
This is done by postulating the instance (there is no special pragma needed).

Agda:
```agda
data Optional (a : Set) : Set where
  Of    : a → Optional a
  Empty : Optional a
{-# COMPILE AGDA2HS Optional #-}

postulate instance iOptionalEq : ⦃ Eq a ⦄ → Eq (Optional a)
{-# COMPILE AGDA2HS iOptionalEq #-}
```

Haskell:
```haskell
data Optional a = Of a
                | Empty

deriving instance (Eq a) => Eq (Optional a)
```

## Deriving Strategies

It is also possible to specify a deriving strategy in the pragma.

NOTE: **deriving via** is not included, since that's quite a bit more complex.

Agda:
```agda
postulate instance iPlanetShow : Show Planet

{-# COMPILE AGDA2HS iPlanetShow derive stock #-}
```

Haskell:
```haskell
deriving stock instance Show Planet
```

### DIFF
--- a/docs/source/features.md
+++ b/docs/source/features.md
@@ -409,7 +409,7 @@ class Ord a where
     x > y = y < x
 ```
 
-### Coppaterns in Type Class Instances
+### Copatterns in Type Class Instances
 
 Agda copatterns are *not* supported by agda2hs in full generality. They *can* be
 used to define fields of typeclass instances, for example:
@@ -436,7 +436,9 @@ instance HasId () where
 
 ### Deriving Type Class Instances
 
-It is also possible to include a standalone `deriving` clause by postulating the instance of a type class.
+It is also possible to include a standalone `deriving` clause by
+* adding the `derive` pragma to an implemented instance,
+* or postulating the instance of a type class.
 
 Agda:
 ```agda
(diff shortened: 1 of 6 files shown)


src/Agda2Hs/Pragma.hs
```diff
@@ -74,12 +81,13 @@ processPragma :: QName -> C ParsedPragma
...
-    | s == "transparent"        -> return TransparentPragma
-    | s == newtypePragma        -> return $ NewTypePragma []
-    | "deriving" `isPrefixOf` s -> processDeriving s DefaultPragma
+    | "class" `isPrefixOf` s     -> return $ ClassPragma (words $ drop 5 s)
+    | s == "existing-class"      -> return ExistingClassPragma
+    | s == "unboxed"             -> return $ UnboxPragma Lazy
+    | s == "unboxed-strict"      -> return $ UnboxPragma Strict
+    | s == "transparent"         -> return TransparentPragma
+    | s == newtypePragma         -> return $ NewTypePragma []
+    | "strategy:" `isPrefixOf` s -> return $ DerivePragma (parseStrategy (drop (length "strategy:") s))
```
flupe, line 90: I'm in favor of dropping the `strategy:` (do you even check "strategy" is written?) as the fact you're compiling an instance makes it redundant.

flupe (commented):
Looking nice!

I'm wondering if it makes sense to generate standalone deriving declarations *only* for *postulated* instances. In particular, if you define a wrapper record (on the Agda side) that will be compiled as a `newtype`, it is very likely you can define a proper instance yourself, even if the Haskell instance ought to be derived automatically.

docs/source/features.md
```diff
@@ -407,7 +409,7 @@ class Ord a where
     x > y = y < x
 ```
 
-## Coppaterns
+### Coppaterns in Type Class Instances
```
jespercockx, line 412: This is not your fault, but it should be "copatterns" (or "co-patterns") rather than coppatterns. No cops allowed here!

jespercockx (approved):
Looks good to me. This is a great improvement to dealing with `deriving`!

src/Agda2Hs/Pragma.hs
sarajuhosova: I've changed this to be included as part of the `derive` clause.

src/Agda2Hs/Compile/ClassInstance.hs
```diff
@@ -42,16 +43,21 @@ enableCopatterns = local $ \e -> e { copatternsEnabled = True }
 disableCopatterns :: C a -> C a
 disableCopatterns = local $ \e -> e { copatternsEnabled = False }
 
-compileInstance :: Definition -> C (Hs.Decl ())
-compileInstance def@Defn{..} = enableCopatterns $ setCurrentRangeQ defName $ do
+compileInstance :: InstanceTarget -> Definition -> C (Hs.Decl ())
+compileInstance ToDerivation def@Defn{..} = do
+  tellExtension StandaloneDeriving
   ir <- compileInstRule [] (unEl defType)
```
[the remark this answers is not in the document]
sarajuhosova: Added!

sarajuhosova:
@lucas, I've added support for compiling implemented as well as postulated instances into a `deriving` clause. This can now be done with the `derive` pragma, which has also replaced the `strategy` annotation.

flupe (approved):
Looks good to me, well done!

PR #162, opened 2023-03-23
Only run the CI when source files are changed

sarajuhosova:
Since the CI is quite expensive to run, this change will limit it to running only when files related to the source code are changed (so not when documentation is).


jespercockx:
Thanks!

PR #164, opened 2023-03-24
[fix #154] Reject Nat constructors in patterns

### LINKED ISSUE #154: Prevent pattern-matching on `Nat`
I now there is already an issue for allowing pattern-matching on `Nat`s: #117.
Since we're approaching `1.0` and the latter seems quite hard to implement properly,
I'm in favor of doing the converse for now: explicitly error out when someone pattern-matches on values of type `Nat`.

Currently it is allowed and produces bogus Haskell code:

```agda
foo : Nat → Nat
foo zero = zero
foo (suc x) = x

{-# COMPILE AGDA2HS foo #-} 
```

```haskell

foo :: Natural -> Natural
foo zero = zero
foo (suc x) = x
```

I'm also conflicted about how things behave: `Nat` being renamed to `Natural` and an import being automatically added. It would make sense to provide a `Natural` type that _doesn't export its constructors_, and isn't part of the `Prelude`, so that using it on the Agda side requires an import.

Any comment?

omelkonian:


jespercockx (approved):

jespercockx:
Thanks!

flupe:
We had some offline discussions with Jesper since the issue was created and thought it would be "better" to forever disallow compiling `Nat`, even when no pattern-matching is used (as `Natural` isn't even part of base, and doesn't have zero and suc constructors), and instead throw a warning `you should define you own Nat datatype instead` so that a proper datatypes gets generated.

I think this makes sense and simplifies what we have to do (no intricate pattern-matching translation, no automatically generated imports that aren't there on the agda code, etc). Perhaps @jesper

jespercockx:
I don't disagree, but Orestis was faster than you with making a PR :)

PR #168, opened 2023-04-06
Fix for issue #142

### LINKED ISSUE #142: Disable pattern-matching on erased argument
Currently, you are allowed to pattern-match on erased arguments if there is only one constructor available. Consider:

```agda
coerce : @0 a ≡ b → a → b
coerce refl x = x
{-# COMPILE AGDA2HS coerce #-}
```

We get the following Haskell code, that does not type-check:

```agda
coerce :: a -> b
coerce x = x
```

For propositional equality specifically, we may want to compile to equality constraints (#140) when the equality argument is an instance, or to Haskell's builtin propositional equality GADT when the equality is *not* erased.
In general though, probably best to disable pattern-matching on erased arguments entirely, especially if our goal is to produce valid Haskell code.

jespercockx:
This fixes #142 by enforcing that there are no dot patterns in the functions we compile to Haskell, even in implicit (type) arguments. To make this work, I also had to add a check that type parameters that appear in the Haskell code are not marked with `@0` (which was already the case for datatypes, but not yet for functions).

### DIFF
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -15,7 +15,7 @@ import qualified Language.Haskell.Exts.Build as Hs
 import Agda.Compiler.Backend
 import Agda.Compiler.Common
 
-import Agda.Syntax.Common ( NamedArg, namedArg )
+import Agda.Syntax.Common
 import Agda.Syntax.Internal
 import Agda.Syntax.Literal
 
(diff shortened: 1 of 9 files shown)


src/Agda2Hs/Compile/Function.hs
```diff
@@ -162,10 +161,24 @@ compilePats :: NAPs -> C [Hs.Pat ()]
...
-    keepPat p = andM
-      [ return $ keepArg p
-      , not <$> isUnboxCopattern (namedArg p)
-      ]
+    keepPat p = do
+      keep <- return (keepArg p) `and2M` (not <$> isUnboxCopattern (namedArg p))
+      when keep $ noAsPatterns $ namedArg p
+      -- We do not allow forced (dot) patterns for non-erased arguments (see issue #142).
+      when (usableModality p && isForcedPat (namedArg p)) $
+        genericDocError =<< "not supported by Agda2Hs: forced (dot) patterns in non-erased positions"
```
flupe, line 169: Can you use the same capitalization of agda2hs (as `Compile/Type.hs:138`) for both error messages?

flupe (approved):
This is a really neat solution!

src/Agda2Hs/Compile/Function.hs
jespercockx: Done!

PR #170, opened 2023-04-12
Some extra proofs + NewType constructor docs

sarajuhosova:
This is a sort of random PR with things that I added while working with Agda2HS, but I thought they might be useful to include.

It includes:

* Documentation (and a test) on how to get the same constructor name in `newtype` as the type name
* A few proofs about `if_then_else_`,  `++` and `||`
* Exports of proofs that were already there but weren't available

### DIFF
--- a/docs/source/features.md
+++ b/docs/source/features.md
@@ -140,7 +140,29 @@ newtype Identity a = MkIdentity{runIdentity :: a}
 newtype Equal a = MkEqual{pair :: (a, a)}
 ```
 
-_Note: Unfortunately, Agda does not allow the constructor name to be the same as the data/record name._
+Unfortunately, Agda does not allow the constructor name to be the same as the data/record name.
+However, it _is_ possible to achieve this with Agda2HS if you do not explicitly add a constructor name to a record definition (this requires the use of `record { ... }` syntax on the Agda side):
+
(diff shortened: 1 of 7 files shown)


docs/source/features.md
```diff
@@ -140,7 +140,29 @@ newtype Identity a = MkIdentity{runIdentity :: a}
 newtype Equal a = MkEqual{pair :: (a, a)}
 ```
 
-_Note: Unfortunately, Agda does not allow the constructor name to be the same as the data/record name._
+Unfortunately, Agda does not allow the constructor name to be the same as the data/record name.
+However, it _is_ possible to achieve this with Agda2HS if you do not explicitly add a constructor name to a record definition (though the notation on the Agda side will be a bit awkward):
```
jespercockx, line 144: "the notation on the Agda side will be a bit awkward" seems a bit too subjective to include in a user manual, perhaps instead we could say "This requires use of the `record { ... }` syntax to construct an element on the Agda side" or something similar.

sarajuhosova: Good point, I've changed it to "(this requires the use of `record { ... }` syntax on the Agda side)".

jespercockx:
Great, thank you for the PR!

PR #173, opened 2023-05-01
Add some list laws

sourceCode4:
Ported some list proofs inspired by Agda's standard library. Also added \ne shorthand and cong2 to equality.


jespercockx (approved):
Thanks for the PR, and sorry for the delay in getting back to you!

PR #183, opened 2023-07-12
Marking condition fields in Num type class as erased

viktorcsimma:
In the definition of the `Num` type class, conditions for the applicability of some functions were not marked as erased; therefore, they got copied into the Haskell file. Now I think I fixed this.

Thank you in advance:)


jespercockx:
Thanks for the PR!

Issue #117, opened 2022-10-21
Support pattern matching on `Natural`?


amesgen:
Suppose we have
```agda
fib : Nat → Nat
fib 0 = 0
fib 1 = 1
fib (suc (suc n)) = fib n + fib (suc n)
{-# COMPILE AGDA2HS fib #-}
```
Right now, this is translated to
```haskell
fib :: Natural -> Natural
fib zero = 0
fib (suc zero) = 1
fib (suc (suc n)) = fib n + fib (suc n)
```
which does not compile ofc. But this could be translated to very similar code which is actually compiling without warnings:
```haskell
{-# LANGUAGE PatternSynonyms #-}
{-# LANGUAGE ViewPatterns #-}

pattern Zero :: Natural
pattern Zero = 0

pattern Suc :: Natural -> Natural
pattern Suc n <- (minusOne -> Just n) where Suc n = n + 1

{-# COMPLETE Zero, Suc #-}

minusOne :: Natural -> Maybe Natural
minusOne n = if n >= 1 then Just (n - 1) else Nothing

fib :: Natural -> Natural
fib Zero = 0
fib (Suc Zero) = 1
fib (Suc (Suc n)) = fib n + fib (Suc n)
```

I think sth like this could be useful (e.g. inductively constructing trees of a certain height), but maybe it is too ad-hoc/magic?

jespercockx:
This might be a nice idea, provided we also give Haskell implementations for the builtin functions on naturals in Agda (see https://github.com/agda/agda/blob/master/src/data/lib/prim/Agda/Builtin/Nat.agda)

jespercockx:
The argument against this is that the goal of agda2hs is to write *Haskell* code in Agda syntax, while this is instead trying to encode an Agda feature into Haskell syntax. So in a sense this is the "wrong" thing to want, since the `Natural` type of Haskell is very rarely used compared to the `Nat` type in Agda.

uncle-betty:
Maybe output the following Haskell code instead? Maybe that would look a little more idiomatic?

```
fib :: Natural -> Natural
fib 0 = 0
fib 1 = 1
fib n' | n' >= 2 = fib (n' - 2) + fib (1 + (n' - 2))
```
I.e.,

  * Matching with a constant natural number is preserved.
  * Matching with the i-th successor of a variable `n` results in a guard of `n' >= i`. And on the RHS, `n` would become `n' - i`.
  * `suc x` on the RHS becomes `1 + x`.

[...]

PR #184, opened 2023-07-12
Translating "zero" constructor of Nat to "0" literal

viktorcsimma:
I've found @uncle-betty's workaround for natural pattern matching at #117 quite useful. To be more precise, this one, when we make an equivalent version where we only pattern-match for zero:

```agda
module Test where

open import Haskell.Prelude
open import Haskell.Prim using (monusNat)

fac3 : Nat → Nat
fac3 0       = 1
fac3 (suc n) = suc n * fac3 n

{-# TERMINATING #-}
fac3' : Nat → Nat
fac3' 0 = 1
fac3' n = n * fac3' (monusNat n 1)

{-# COMPILE AGDA2HS fac3' #-}

fac3'≡fac3 : ∀ n → fac3' n ≡ fac3 n
fac3'≡fac3 zero = refl
fac3'≡fac3 (suc n) rewrite fac3'≡fac3 n = refl
```

However, this is not supported anymore per #154. I consider it to be useful; so I've rewritten `Function.hs` so that it always rewrites `Agda.Builtin.Nat.Nat.zero` to a `0` literal. Pattern matching to `suc` would still be rejected.

I don't know whether it would require a broader consensus; but I think we wouldn't lose anything by adding this feature. Thanks in advance for looking at it:)

### DIFF
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -43,13 +43,14 @@ isSpecialPat qn = case prettyShow qn of
   "Haskell.Prim.Tuple._;_" -> Just tuplePat
   "Agda.Builtin.Int.Int.pos" -> Just posIntPat
   "Agda.Builtin.Int.Int.negsuc" -> Just negSucIntPat
+  "Agda.Builtin.Nat.Nat.zero" -> Just $ \ _ _ _ -> return $ Hs.PLit () (Hs.Signless ()) (Hs.Int () 0 "0")
   s | s `elem` badConstructors -> Just $ \ _ _ _ -> genericDocError =<<
     "constructor `" <> prettyTCM qn <> "` not supported in patterns"
   _ -> Nothing
   where
     badConstructors =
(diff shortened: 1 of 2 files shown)


jespercockx:
Thanks for the PR, but I am still not convinced that using Agda's builtin `Nat` type in non-erased parts of Agda2Hs makes any sense. There is no default type in Haskell that corresponds to it, so I don't even know what kind of useful code you'd want to get out of this.

viktorcsimma:
You are right about the Natural type… but still, recursive functions with a case for 0 and a case for everything else are legal Haskell code, maybe even idiomatic. (A factorial function is something I can imagine as an example – that could be written in Agda with integers and an erased proof that they are nonnegative, but it would look strange.) Maybe the Natural type can be replaced with Integer, but we should think about how to handle the negative cases then… I don’t what the best solution would be for that. But I think that might belong to a separate issue.

viktorcsimma:
Maybe an idea for a replacement of naturals would be some kind of record, with an integer and an erased proof. And that could be used in the Agda code, and the Haskell code would simply work with Integers.

PR #188, opened 2023-07-16
Add and improve documenation

wcqaguxa:
- Fixed a few typos in existing documentation
- Uploading first two simple tutorials. The example projects they link to are still hosted on my personal github account -should they be also moved into this repository? If so, where do they fit, structurally?


docs/source/tutorials.md
```diff
@@ -0,0 +1,103 @@
...
+```agda
+open import Definition
+```
+It does not have to be specified for agda2hs to compile this import. If the import is used by any of the compiled code, the relevant Haskell module will be likewise imported in the target file; the irrelevant imports are automatically skipped. 
+
+However, both files need to import `Haskell.Prelude` independently of each other to be able to use it.
+
+### Compilation
+
+In principle, `agda2hs` has to be invoked on file-by-file basis and to define a compilation of a whole folder, additional scripting is neccesary. However, since `Definition` is imported in `Usage`, when compiling the latter, the former also will be compiled as a ncessary dependency. Thus, in our example, it is only necessary to execute `agda2hs` once.
```
jespercockx, line 97: Typo: `ncessary`

jespercockx:
Thanks for the PR! I think we could add the tutorials to a `tutorials` subdirectory in this repo.

tutorials/example-basics/HelloWorld.agda
jespercockx, line 1: minor nitpick: please make sure that each file ends with a newline (git doesn't like it otherwise)

PR #189, opened 2023-07-18
Option for user-defined rewrite rules given in a config file

viktorcsimma:
The rewriting mechanism used in `isSpecialName` in `Name.hs` has been opened for users. For this, you have to use the `--rewrite-rules` option and provide a YAML file with the format of `rewrite-rules-example.yaml`.

Why is this useful? Imagine you have a project which uses the Agda standard library quite extensively. You don't want to rewrite the entire standard library to an agda2hs-compatible version, but you need the theorems contained in it for your proofs. With this option, you can translate the operators of the standard library to the corresponding Haskell functions, or to your predefined functions (I used a file called `Base.hs` for that in my project).

Of course, that compromises the mathematically proven correctness of your library a bit. But if you trust that the Haskell Prelude operators do the same thing as the Agda standard library functions, this might not be something to worry about. If you want, you can even write proofs for the equivalence of the two. Besides that, the feature would remain optional.

Known issues:
- If you import something from Prelude, you have to state this explicitly.
- Then, it will add an import like `import Prelude (...)`, which shadows all the other functions in Prelude. I think this is not necessarily a problem; as we will probably not modify the generated Haskell files by hand, and so we won't add any new Prelude functions.
- But this only works for things you do use, not for those that you only define. For example, if you write an instance of the `Num` class and define `signum` but don't use it, it won't get into Prelude's import list, and so GHC will complain.
- You cannot change to a version with arguments of different modality without getting rubbish code. So if you rewrite a function to a version which has some of its parameters erased, the parameters remain there; probably because rewriting happens only after compiling the type signature.

Further development:
- I would like to add a flag which gives a warning for all names defined neither in the project nor in a rewrite rule. This would make purging invalid references easier.
- Maybe an implicit import of Prelude functions. But with that, you could get into trouble if you reuse a name already defined in Prelude.

### DIFF
--- a/agda2hs.cabal
+++ b/agda2hs.cabal
@@ -56,7 +56,7 @@ executable agda2hs
                        AgdaInternals
   build-depends:       base >= 4.10 && < 4.18,
                        Agda >= 2.6.3 && < 2.6.4,
-											 bytestring >= 0.11,
+                       bytestring >= 0.11,
                        containers >= 0.6 && < 0.7,
                        unordered-containers >= 0.2,
                        mtl >= 2.2,
@@ -66,7 +66,7 @@ executable agda2hs
                        syb >= 0.7,
                        text >= 1.2.3.0,
                        deepseq >= 1.4.1.1,
-											 yaml-light >= 0.1.4
+                       yaml-light >= 0.1.4
   default-language:    Haskell2010
   default-extensions:  LambdaCase,
                        RecordWildCards,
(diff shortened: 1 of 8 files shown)


jespercockx:
This is a very nice and useful feature to have! Could you please add a short explanation to the user manual with a quick example?

viktorcsimma:
Done; thank you:)

jespercockx:
I merged it! I saw there are still a couple of open issues listed in the documentation, perhaps you could create new issues for those here on github as well, so other people who might want to help can jump in?

viktorcsimma:
Of course:) Thanks again!

PR #193, opened 2023-07-21
Rewrite rules: specifiable imports from Prelude

### LINKED ISSUE #192: Rewrite rules' handling of Prelude (and a bug when instantiating a Haskell type class)
Currently, the rewrite rules feature handles Prelude the following way:
- If you import something from Prelude, you have to state this explicitly. (This can mean a _lot_ of inconvenient rewrite rules.)
- Then, it will add an import like import Prelude (...), which shadows all the other functions in Prelude. I think this is not necessarily a problem; as we will probably not modify the generated Haskell files by hand, and so we won't add any new Prelude functions.
- But this only works for things you do use, not for those that you only define. For example, if you write an instance of the Num class and define signum but don't use it, it won't get into Prelude's import list, and so GHC will complain.

There could be multiple solutions for this.
- Make imports from Prelude implicit (don't add an ˙import Prelude (...)` line). This would solve the last problem quickly, but would cause trouble if we want to reuse a name from Prelude.
- In instances, apply the rewrite rule to left sides, too; or add them to Prelude's import list automatically.
- Provide the user an option about whether they want to import Prelude functions implicitly.

See #189 for the current state of related code.

viktorcsimma:
Closes #192. Adds the opportunity to define the handling of Prelude in the config file. See the documentation update for details:

By default, agda2hs handles Prelude like other modules: it collects all the identifiers it finds we use from Prelude, and adds them to Prelude's import list.

In the config YAML, we can specify a different behaviour. The format is:

```yaml
# First, we specify how to handle Prelude.
prelude:
  implicit: true
  hiding:           # if implicit is true
    - seq

  #using:           # if implicit is false
  #  - +
  #  - Num

# Then the rules themselves.
rules:

  # The rational type.
  - from: Data.Rational.Unnormalised.Base.ℚᵘ
    to: Rational
    importing: Data.Ratio
  - [...]
```

If `implicit` is `true`, then everything gets imported from Prelude, except for those that are specified in the `hiding` list. This can cause clashes if you reuse names from Prelude, hence the opportunity for a `hiding` list. If there is no such list, then everything gets imported.

If `implicit` is `false`, Prelude gets imported explicitly, and only those identifiers that are specified in the `using` list. If there is no such list, agda2hs reverts to the default behaviour (it tries to collect imports by itself).

### DIFF
--- a/docs/source/features.md
+++ b/docs/source/features.md
@@ -891,7 +891,9 @@ testA = A.foo
 testB = A.foo
 ```
 
-# Rewrite rules
+# Rewrite rules and Prelude imports
+
+## Rewrite rules
 
 User-defined rewrite rules for names can be defined through YAML configuration files. These should be provided via the `--rewrite-rules` option.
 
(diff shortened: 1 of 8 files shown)


jespercockx:
Thanks, this is great once again! I'm actually thinking that perhaps it would even be a good idea to have implicit prelude set to `true`, since it is less likely to cause problems with identifiers being out of scope (at the cost of potentially having name clashes, but it is probably a good thing to have an error when you accidentally introduce a name clash with a function from Prelude).

viktorcsimma:
Thank you:) I've tried out setting the default to an implicit Prelude, but this broke the tests in cases where `Haskell.Prelude` was imported with a qualifier (usually `Pre`). This might be solved by a qualified import, but then we would have to deal with the case when some of it is imported with a qualifier and some of it is not. Another option would be to simply delete the qualifier, but this would then create clashes that did not exist in the Agda code.

What do you think? Maybe we should leave it as it is for now and create an issue for changing the default.

jespercockx:
Sure, that's a good idea.

PR #194, opened 2023-07-24
Adding "type" prefix before type operator imports

viktorcsimma:
GHC expects that when importing type operators, the `type` prefix is provided before the identifier in the import specs. This PR fixes that. It only calculates this for operators; since for other identifiers it would only be needed when having a "normal" function with the same name, and Agda doesn't allow that.

Unfortunately, when type operators have "children" imports (see `(***)` in `TypeOperatorImport.hs`), you have to specify it too, but the Haskell Language Extensions' pretty printer does not do it. So I had to circumvent this; which has the side effect of it not being able to break long import lines anymore. (But see [PR 475](https://github.com/haskell-suite/haskell-src-exts/pull/475) there.)

### DIFF
--- a/src/Agda2Hs/Compile/Imports.hs
+++ b/src/Agda2Hs/Compile/Imports.hs
@@ -20,7 +20,9 @@ import Agda2Hs.Compile.Types
 import Agda2Hs.Compile.Utils
 import Agda2Hs.HsUtils
 
-type ImportSpecMap = Map (Hs.Name ()) (Set (Hs.Name ()))
+type MaybeNamespace = Maybe (Hs.Namespace ()) -- just for shortening
+
+type ImportSpecMap = Map (Hs.Name (), MaybeNamespace) (Set (Hs.Name (), MaybeNamespace))
 type ImportDeclMap = Map (Hs.ModuleName (), Qualifier) ImportSpecMap
 
 compileImports :: String -> Imports -> TCM [Hs.ImportDecl ()]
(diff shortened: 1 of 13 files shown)


src/Agda2Hs/Compile/Imports.hs
```diff
@@ -20,7 +20,7 @@ import Agda2Hs.Compile.Types
 import Agda2Hs.Compile.Utils
 import Agda2Hs.HsUtils
 
-type ImportSpecMap = Map (Hs.Name ()) (Set (Hs.Name ()))
+type ImportSpecMap = Map (Hs.Name (), Hs.Namespace ()) (Set (Hs.Name (), Hs.Namespace ()))
```
jespercockx, line 23: Perhaps you could introduce a new type that packages a `Name` and a `Namespace`? These tuples are a bit too easy to get confused about.

src/Agda2Hs/Compile/Name.hs
```diff
@@ -147,15 +155,37 @@ compileQName f
...
+                                              ++ "is a type operator: " ++ show isQNameSet
+      if isQNameSet then return (Hs.TypeNamespace ()) else return (Hs.NoNamespace ())
+
+    -- Gets the type of the result of the function (the type after the last "->").
+    getResultType :: Type -> Type
+    getResultType typ = case (unEl typ) of
+      (Pi _ absType) -> getResultType $ unAbs absType
+      _              -> typ
+
+    -- I'm not sure whether this is the correct way to determine whether it's a Set/Prop;
```
jespercockx, line 174: this does not seem right, you should instead look at the *argument* of the `Sort` constructor to check whether it is a `Set` sort or `Prop` sort. Or even better, use the `isPropM` utility function.

jespercockx (approved):
Thank you for the PR! I have two comments, but overall this seems like a useful improvement to have.

src/Agda2Hs/Compile/Name.hs
viktorcsimma: Maybe I don't understand something about the concepts themselves. I've tried `isQNameSet <- isPropM $ getResultType $ defType definition` at line 164, but then I got the same strange results as when I looked at levels: in `TypeOperatorImport.hs`, I've got

```hs
import TypeOperatorExport ((&&&), (***), type (***)((:*:)), (<))
```

The sorts themselves were strange. `(:*:)` seemed to have the same sort (`InfS ...`) as `(***)`. And there were type operators with level `Max 0 ...` and normal operators `Max 1 ...`, and then vice versa. Is there something from the theory which I don't understand? Thank you in advance:)

src/Agda2Hs/Compile/Imports.hs
viktorcsimma: This is done now:)

src/Agda2Hs/Compile/Name.hs
jespercockx: > The sorts themselves were strange. (:*:) seemed to have the same sort (InfS ...) as (***). And there were type operators with level Max 0 ... and normal operators Max 1 ..., and then vice versa. Is there something from the theory which I don't understand? Thank you in advance:)

The reason why these sorts are `InfS` is because both symbols are universe-polymorphic:
```agda
data _***_ {i j : Level} (a : Set i) (b : Set j) : Set (i ⊔ j) where
  _:*:_ : a -> b -> a *** b
```
so they both end up in the sort of universe-polymorphic things, which is `InfS ...`. This suggests that just looking at the sort is not really the right thing to do here. I think that you could instead look at the return type of the symbol and check whether it is a sort (`Sort` term constructor) or a datatype (`Def` term constructor). Could you give that a try and let me know how it goes?

viktorcsimma: Thank you:) If I understand well, I think this was my original concept: `getResultType` gets the return type through a recursion and `isSet` decides based on the constructor whether it's a sort. What I'm unsure about is how to handle the other constructors (`Lam`, `Lit` etc.); probably there is no need for type namespace for them.

jespercockx: Ah yes, I see now what you were trying to do. In that case I think your implementation is actually fine, I was just confused by the name `isSet` which should more properly be called `isSort`. And actually this helper function already exists in `Agda.Syntax.Internal` (though it returns a `Maybe Sort` instead of a `Bool`).

viktorcsimma: I didn't know about that; thanks! I'm using that now.

PR #195, opened 2023-07-26
Skipping erased value arguments inside type functions in datatype parameters and in type lambdas

viktorcsimma:
When giving type parameters to a record and using a dependent function which, however, has its value parameters erased; I've originally got an error (`Kind of bound argument not supported`). This PR fixes that.

The specific example (used in the test case, too):

```agda
open import Agda.Primitive

-- A record type which has both members compiled,
-- but the argument of the lambda is erased;
-- so that it won't be dependent-typed after compilation.
record Σ' {i j} (a : Set i) (b : @0 a → Set j) : Set (i ⊔ j) where
  constructor _:^:_
  field
    proj₁ : a
    proj₂ : b proj₁
open Σ' public
infixr 4 _:^:_
{-# COMPILE AGDA2HS Σ' #-}
```

The output here is

```hs
data Σ' a b = (:^:){proj₁ :: a, proj₂ :: b}
```

**Update:** now the arguments of type lambdas are also checked and if they are erased, they are dropped, instead of invoking an error. An example:

```agda
-- Actually, Agda can deduce here that n is erased; probably from the type signature of Σ'.
test : Nat -> Σ' Nat (λ (n : Nat) -> ⊤)
test n = n :^: tt
{-# COMPILE AGDA2HS test #-}
```

This produces an output like this:

```hs
test :: Natural -> Σ' Natural ()
test n = n :^: ()
```

### DIFF
--- a/src/Agda2Hs/Compile/Type.hs
+++ b/src/Agda2Hs/Compile/Type.hs
@@ -193,5 +193,7 @@ compileKeptTeleBind x t = do
 compileKind :: Type -> Maybe (Hs.Kind ())
 compileKind t = case unEl t of
   Sort (Type _) -> pure (Hs.TyStar ())
-  Pi a b -> Hs.TyFun () <$> compileKind (unDom a) <*> compileKind (unAbs b)
-  _ -> Nothing
+  Pi a b
+    | usableModality a  -> Hs.TyFun () <$> compileKind (unDom a) <*> compileKind (unAbs b)
+    | otherwise         -> compileKind (unAbs b)
+  _ -> Nothing          -- ^ if the argument is erased, we only compile the rest
--- a/src/Agda2Hs/Compile/Type.hs
+++ b/src/Agda2Hs/Compile/Type.hs
@@ -141,7 +141,7 @@ compileType t = do
       return $ tApp (Hs.TyVar () x) vs
     Sort s -> return (Hs.TyStar ())
     Lam argInfo restAbs
-      | not (keepArg argInfo)   -> compileType $ unAbs restAbs
+      | not (keepArg argInfo)   -> underAbstraction_ restAbs compileType
     _ -> genericDocError =<< text "Bad Haskell type:" <?> prettyTCM t
 
 compileTypeArgs :: Args -> C [Hs.Type ()]
@@ -186,16 +186,21 @@ compileTeleBinds tel =
 compileKeptTeleBind :: Hs.Name () -> Type -> C (Hs.TyVarBind ())
 compileKeptTeleBind x t = do
   checkValidTyVarName x
-  case compileKind t of
+  compileKind t >>= (\case
     Just k              -> pure $ Hs.UnkindedVar () x -- In the future we may want to show kind annotations
     _                   -> genericDocError =<<
       text "Kind of bound argument not supported:"
-      <+> parens (text (Hs.prettyPrint x) <> text " : " <> prettyTCM t)
+      <+> parens (text (Hs.prettyPrint x) <> text " : " <> prettyTCM t))
 
-compileKind :: Type -> Maybe (Hs.Kind ())
+compileKind :: Type -> C (Maybe (Hs.Kind ()))
 compileKind t = case unEl t of
-  Sort (Type _) -> pure (Hs.TyStar ())
-  Pi a b
-    | keepArg a    -> Hs.TyFun () <$> compileKind (unDom a) <*> compileKind (unAbs b)
-    | otherwise    -> compileKind (unAbs b)
-  _ -> Nothing     -- ^ if the argument is erased, we only compile the rest
+  Sort (Type _) -> return (Just $ Hs.TyStar ())
+  Pi a b -> (compileKind $ unDom a) >>= (\case
+    Just kindA -> do
+      mbKindB <- underAbstraction a b compileKind
+      return (Hs.TyFun() kindA <$> mbKindB)
+    Nothing
+      | keepArg a    -> return Nothing
+      | otherwise    -> underAbstraction a b compileKind)
+                     -- ^ if the argument is erased, we don't complain about it
+  _ -> return Nothing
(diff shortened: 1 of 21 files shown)


viktorcsimma:
It now seems like that it compiles, but cannot be used... I'll try to fix that.

viktorcsimma:
Now it's OK:) I've inserted a similar check at type lambdas. See the update above.

src/Agda2Hs/Compile/Type.hs
```diff
@@ -140,7 +140,11 @@ compileType t = do
       x  <- hsName <$> compileVar x
       return $ tApp (Hs.TyVar () x) vs
     Sort s -> return (Hs.TyStar ())
-    t -> genericDocError =<< text "Bad Haskell type:" <?> prettyTCM t
+    Lam argInfo restAbs
+      | usableModality argInfo  -> genericDocError =<< text "Bad Haskell type:" <?> prettyTCM t
```
jespercockx, line 144: You can leave out this line and just let it fall through to the `_ ->` case below

jespercockx: Also, please use the `keepArg` utility function (https://github.com/agda/agda2hs/blob/fcd2885e3775cfb96cf100303510f7485fb5e82c/src/Agda2Hs/Compile/Utils.hs#L120) instead of `usableModality`.

src/Agda2Hs/Compile/Type.hs
```diff
@@ -193,5 +197,7 @@ compileKeptTeleBind x t = do
 compileKind :: Type -> Maybe (Hs.Kind ())
 compileKind t = case unEl t of
   Sort (Type _) -> pure (Hs.TyStar ())
-  Pi a b -> Hs.TyFun () <$> compileKind (unDom a) <*> compileKind (unAbs b)
-  _ -> Nothing
+  Pi a b
+    | usableModality a  -> Hs.TyFun () <$> compileKind (unDom a) <*> compileKind (unAbs b)
```
jespercockx, line 201: Same here: use `keepArg` and `underAbstraction`

src/Agda2Hs/Compile/Type.hs
```diff
@@ -140,7 +140,11 @@ compileType t = do
       x  <- hsName <$> compileVar x
       return $ tApp (Hs.TyVar () x) vs
     Sort s -> return (Hs.TyStar ())
-    t -> genericDocError =<< text "Bad Haskell type:" <?> prettyTCM t
+    Lam argInfo restAbs
+      | usableModality argInfo  -> genericDocError =<< text "Bad Haskell type:" <?> prettyTCM t
+      | otherwise               -> compileType $ unAbs restAbs
```
jespercockx, line 145: Use the `underAbstraction` utility function to make sure the TCEnv is updated properly (see the `Pi` case above).

src/Agda2Hs/Compile/Type.hs
viktorcsimma: Done:)

src/Agda2Hs/Compile/Type.hs
viktorcsimma: Here, I'm not sure we can do it: the first argument of `Lam` is only an `ArgInfo`, not a `Dom` as in `Pi`. I've tried to look for functions which create a `Dom Type` for that, but I couldn't find any. Or is `defaultArgDom` for that? But then, we would need a `Type`, which we still don't have for that part.

src/Agda2Hs/Compile/Type.hs
viktorcsimma: Oh, sorry; I've missed that... I'll do that soon; this is a `Pi`, so I think I won't have problems with it.

viktorcsimma:
One more thing apart from the comments: `Type.hs` has some more references to `usableModality`; should we change them to `keepArg` too? Or is there a reason why they are used there?

src/Agda2Hs/Compile/Type.hs
jespercockx: You can use `underAbstraction_`, which creates a dummy domain (not perfect, but at least then the de Bruijn indices are not all messed up)

jespercockx:
> One more thing apart from the comments: Type.hs has some more references to usableModality; should we change them to keepArg too? Or is there a reason why they are used there?

For types it is a bit more tricky since there is the possibility of an instance argument. These can be erased at the term level, but at the type level they need to be made into a constraint. So I don't think anything there should change.

src/Agda2Hs/Compile/Type.hs
viktorcsimma: I think it's done now:) I hope I understood it well.

src/Agda2Hs/Compile/Type.hs
viktorcsimma: That was what I was looking for:)

src/Agda2Hs/Compile/Type.hs
```diff
@@ -184,14 +186,21 @@ compileTeleBinds tel =
 compileKeptTeleBind :: Hs.Name () -> Type -> C (Hs.TyVarBind ())
 compileKeptTeleBind x t = do
   checkValidTyVarName x
-  case compileKind t of
+  compileKind t >>= (\case
```
jespercockx, line 189: Small nitpick: you don't need to have parentheses around the `\case ...` here. In general it's good style to avoid parentheses that span multiple lines.

src/Agda2Hs/Compile/Type.hs
```diff
@@ -184,14 +186,21 @@ compileTeleBinds tel =
...
     Just k              -> pure $ Hs.UnkindedVar () x -- In the future we may want to show kind annotations
     _                   -> genericDocError =<<
       text "Kind of bound argument not supported:"
-      <+> parens (text (Hs.prettyPrint x) <> text " : " <> prettyTCM t)
+      <+> parens (text (Hs.prettyPrint x) <> text " : " <> prettyTCM t))
 
-compileKind :: Type -> Maybe (Hs.Kind ())
+compileKind :: Type -> C (Maybe (Hs.Kind ()))
```
jespercockx, line 195: Actually I am stupid: `compileKind` was not a monadic function so there's no context to worry about. So using `underAbstraction` here is not really necessary after all. I'm sorry for the extra work, but could you make it pure again?

jespercockx: Thanks!

PR #196, opened 2023-08-06
Update to Agda 2.6.4

plt-amy:
On the Haskell side, other than me having moved `Agda.Utils.Pretty` (possibly the most breaking a trivial change could be), the values in `iForeignCode` are now also wrapped in a newtype.
On the Agda side, `--erasure` is now a flag; I added it to the agda-lib files rather than adding it by hand everywhere.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
       if: github.event.action == 'opened' || github.event.action == 'synchronize' || github.event.ref == 'refs/heads/master'
 
     # Takes care of ghc and cabal. See https://github.com/haskell/actions.
-    - uses: haskell/actions/setup@v2
+    - uses: haskell-actions/setup@v2
       id: setup-haskell-cabal
       name: Setup Haskell
       with:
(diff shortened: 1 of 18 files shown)


jespercockx:
Anyone knows how to force CI to re-run `cabal update`? It seems it is currently using a cached version that doesn't have Agda 2.6.4.

omelkonian:
> Anyone knows how to force CI to re-run `cabal update`? It seems it is currently using a cached version that doesn't have Agda 2.6.4.

There seems to be no cache on Github: https://github.com/agda/agda2hs/actions/caches, so probably an upstream cache issue. I see in the logs we are still using the deprecated setup-haskell action:
```
Warning: As of 2023-09-09, haskell/action/setup is no longer maintained, please switch to haskell-actions/setup (note: dash for slash).
```
Let's see if changing this solves the issue..

omelkonian:
Ha, found the culprit by printing `agda2hs.cabal` in the CI log:
```
...
executable agda2hs-mode
  ...
  build-depends:       ..., Agda >= 2.6.3 && < 2.6.4, ...
```
I've rebased on HEAD and changed this as well, now it should work!

omelkonian:
After rebasing, everything seems to work, I will go ahead and merge.

PR #197, opened 2023-08-07
Add postulated IO bindings (#180)

### LINKED ISSUE #180: Add bindings for the IO monad
I just realized we do not have bindings for the IO monad in the `agda2hs` Prelude. Obviously it's not possible to implement them on the Agda side, but we should have postulates for the common IO operations in the Prelude as well as the (lawful) instances for Functor, Applicative, Monad, etc.

Cogsad:
This fixes #180 

* Adds  postulated types(e.g the IO type and the lawful instances for IO) and operations for the IO Monad.

* Adds two test cases for IO operations for file handling and input handling.


jespercockx (approved):
This is great, thank you for the PR!

PR #198, opened 2023-08-10
Adding Emacs mode

viktorcsimma:
Through this, Emacs can interact with the Agda interpreter built into the executable. For now, it can:
- load files, type-check them etc. – anything that you can do with a normal Agda mode;
- compile files using the agda2hs backend (after `C-c C-x C-c`, type `agda2hs`);
- compile with any other usual backend Agda has (GHC, HTML etc.).

Installation is done by inserting the following into `.emacs` (assuming `agda2hs-mode` is under `~/.cabal/bin`):

```el
(load-file (let ((coding-system-for-read 'utf-8))
                (shell-command-to-string "~/.cabal/bin/agda2hs-mode locate")))
```

`agda2hs-mode`'s `Main.hs` is based on Agda 2.6.3's `src/agda-mode`; the Lisp files on `src/data/emacs-mode`. I've only done some minor modifications to them.

Issues:
- Now, `agda2-mode.el` assumes the executable is at `~/.cabal/bin/agda2hs`. I've tried with simply `agda2hs` (as the original Agda mode works), but it doesn't find it this way, even if it's in `PATH` on the terminal. (Maybe Emacs has a different path variable. But it finds the normal `agda` executable; so I don't quite understand.)
- Everything is named `Agda` and `agda2`; so there is no implication that this is not a normal Agda backend. But I think this can be easily fixed by rewriting a few strings.

I'm currently rewriting a large library to be agda2hs-compatible, and this Emacs mode is going to save me hours as:
- I don't have to always type-check from the start even if I've only rewritten a few lines; and
- I can easily do everything from Emacs, even compilation.

What do you think? The path issue should probably be fixed before merging; I'll make some research about it.

### DIFF
--- a/agda2hs.cabal
+++ b/agda2hs.cabal
@@ -15,6 +15,9 @@ description:
   The tool is implemented as an Agda backend, which means that `agda2hs` is a
   fully functional Agda compiler.
 
+data-dir:           src/data
+data-files:         emacs-mode/*.el
+
 source-repository head
   type:     git
   location: https://github.com/agda/agda2hs.git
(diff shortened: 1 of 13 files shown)


jespercockx:
This is great, thank you for the PR! I agree that the path issue should ideally be fixed before merging, IIRC Emacs indeed uses a different path than the default one.

> Switching versions doesn't work yet (which is complicated by the different version numbers of Agda and agda2hs).

Would there be some way to change the version numbering of `agda2hs` to make this easier? Perhaps we could start each version number with the Agda version number?

This also made me wonder whether there should be some way to indicate that an Agda file was written with `agda2hs` in mind, so the mode could be activated automatically. But this could also easily be a separate PR.

> Maybe we should write Haskell files to a new directory called agda2hs/ (like the HTML backend does).

Yes this sounds like a good idea, but probably should be a separate PR as well.

viktorcsimma:
We now add `~/.cabal/bin` to `PATH` in `.emacs`. This way, it should find the executable either here or under the usual directories in Emacs' `PATH` variable.

viktorcsimma:
I've forgotten the version numbers: as I can see, we don't append it to the executable name yet. I'll take a look at how Agda solves this.

viktorcsimma:
And thank you for the other ideas too:) I think I'll open issues for them after merging.

viktorcsimma:
But maybe I would leave the version number issue to a separate PR, too – I think it'll be a long run.

jespercockx:
Ok, so is this PR ready to be merged then?

viktorcsimma:
I think yes, if you agree:) maybe the documentation which should be done before. I’ll do that before merging.

viktorcsimma:
Done;)

PR #199, opened 2023-08-26
Proofs tutorial

wcqaguxa:
The tutorial describes how to get started up with proving things. I'm not 100% sure if everything described here is factually correct (word choices and things described as impossible / not working) and if the granularity is correct  (too much or too little explanations on certain aspects? )


docs/source/tutorials.md
```diff
@@ -101,3 +101,197 @@ Since the target repository is different than the source, it has to be specified
...
+
+[Example source code](https://github.com/agda/agda2hs/tree/master/tutorials/example-proofs)
+
+This tutorial aims to explain how to apply different formal verification techniques compliant with agda2hs and haskell. 
+
+### Constructor constraints
+
+The code described in this section can be found in the file `Triangle.agda`.
+
+Lets say we want to have a data type describing a triangle. A first attempt might look somewhat like this:
```
jespercockx, line 114: Grammar: "let's"

docs/source/tutorials.md
```diff
@@ -101,3 +101,197 @@ Since the target repository is different than the source, it has to be specified
...
+```agda
+open import Haskell.Prelude
+
+data Triangle : Set where
+    MkTriangle : (alpha beta gamma : Nat)
+           → Triangle
+
+{-# COMPILE AGDA2HS Triangle #-}
+```
+It's defined with three angles, because maybe thats the only property we are interested in so far. However, three arbitrary angle values do not make a triangle. First of all, they cannot be negative - to make things easier we use natural numbers (`Nat`) to represent the angle values. But that is not enough, three angles can constitute a triangle only if they sum up to 180 degrees and at most one angle is right or obtuse. This can be modeled in Agda:
```
jespercockx, line 125: Isn't it enough to have all three angles be strictly larger than 0 and sum up to 180?

docs/source/tutorials.md
```diff
@@ -101,3 +101,197 @@ Since the target repository is different than the source, it has to be specified
...
+
+```agda
+isAscending : ⦃ iOrdA : Ord a ⦄ → List a → Bool
+isAscending [] = True
+isAscending (x ∷ []) = True
+isAscending (x ∷ y ∷ xs) = if x <= y then isAscending (y ∷ xs) else False
+
+{-# COMPILE AGDA2HS isAscending #-}
+```
+This function can be compiled to Haskell without any issue, however, when you try using it in proofs you can notice that it is not the most handy definition: the different cases are anonymous and to invoke them, some heavy stretching with regards to values has to be invoked. A better definition might be a predicate, instead:
```
jespercockx, line 198: > some heavy stretching with regards to values has to be invoked

I'm not sure what you mean by this, could you clarify?

docs/source/tutorials.md
```diff
@@ -101,3 +101,197 @@ Since the target repository is different than the source, it has to be specified
...
+```agda
+data IsAscending₂ {a : Set} ⦃ iOrdA : Ord a ⦄ : List a → Set where
+    Empty : IsAscending₂ []
+    OneElem : (x : a) →  IsAscending₂ (x ∷ [])
+    ManyElem : (x y : a) (xs : List a) 
+        → ⦃ IsAscending₂ (y ∷ xs) ⦄
+        → ⦃ IsTrue (x <= y) ⦄
+        → IsAscending₂ (x ∷ y ∷ xs)
+```
+However, this data type cannot be compiled to Haskell, as they do not allow to match on specific values. However, some amount of equivalency can be proven between these two definitions. They are not structurally different (one evaluates to True and the other to a data type), therefore they cannot be equal per the `≡` operator, however there occurs the material equivalence relation (`⇔`), such that if the function returns `True`, the predicate holds, and vice versa. Lets try to prove it then!
```
jespercockx, line 209: Style: Avoid starting several several sentences with "however" one after another.

docs/source/tutorials.md
```diff
@@ -101,3 +101,197 @@ Since the target repository is different than the source, it has to be specified
...
+helper₁ x xs h₁ with (isAscending (x ∷ xs)) in h₂
+helper₁ x (x₁ ∷ xs) h₁ | True  with (x <= x₁)
+helper₁ x (x₁ ∷ xs) h₁ | True | True = magic (absurd₁ h₂ h₁)
+helper₁ x (x₁ ∷ xs) h₁ | True | False = sym h₂
+helper₁ x xs h₁ | False = refl
+```
+Here is a helper method for the inductive hypothesis. Notice that where in the predicate syntax we were able to pattern-match on the different constructor, when working with a function, the only way to narrow down to different cases is to pattern match on the possible values of the function output. The [`with ... in` syntax](https://agda.readthedocs.io/en/latest/language/with-abstraction.html) can be used in such cases. In the first useage of the syntax, the output of the function is needed to be applied in the proof, so it needs to be saved *in* a value to be accessed in the context. This is what applying `in h₂` achieves. In the second usage, Agda manages to simplify the necessary arguments automatically, so it is not necessary to add the assertion to the syntax. 
+
+The `absurd₁` function can be applied to the `magic` function to resolve the internal contradiction. `magic` lives in `Haskell.Prim` which also needs to be imported.
```
jespercockx, line 272: It's the other way around: the `magic` function is applied to the result of the `absurd` function.

docs/source/tutorials.md
```diff
@@ -101,3 +101,197 @@ Since the target repository is different than the source, it has to be specified
...
+
+The `absurd₁` function can be applied to the `magic` function to resolve the internal contradiction. `magic` lives in `Haskell.Prim` which also needs to be imported. 
+
+```agda
+--recursion helper
+postulate
+   theorem₂helper : ⦃ iOrdA : Ord a ⦄ (xs : List a) → (IsTrue (isAscending xs)) → IsAscending₂ xs
+```
+The last element is a definition of the same type as the actual proof. This was incidentally necessary as the termination check did not recognize that it is being applied to a recursive case. [Postulates](https://agda.readthedocs.io/en/latest/language/postulates.html) are in general a bad practice.
```
jespercockx, line 279: Isn't there some way to get rid of this postulate?

jespercockx (approved):
Thanks for the PR! I have a few small comments but in general this is looks very nice and useful for people new to Agda and `agda2hs`.

docs/source/tutorials.md
wcqaguxa: This depends on if we want to include degenerate triangles in our definition of valid triangles, as they are the only edge case difference between these two given definitions. Should I elaborate about this in the tutorial? I thought it might be a bit off-topic to go into great details of geometry definitions.

docs/source/tutorials.md
wcqaguxa: Actually there is. I suppose I needed a fresh look at the code. I decided to leave the version with postulate to show that it should be avoided, and also because the helper functions written for this version use some concepts that I wanted to cover in the tutorial.

docs/source/tutorials.md
wcqaguxa: Is the new version better?

docs/source/tutorials.md
jespercockx: Probably not worth going into any details here, but perhaps you could consider just using the simple definition.

jespercockx:
Thanks for the updates! I think it's in a good state now so I'll merge it, feel free to make changes later if you notice opportunities for improvement.

PR #205, opened 2023-10-12
Issue 185

### LINKED ISSUE #185: Omitted parameters of functions declared within record namespace
As I can see, functions declared within record namespace often lose their parameters; maybe because their type is evaluated wrong.

Here is my example code:

```agda
open import Agda.Builtin.Bool

record RecordTest : Set where
  constructor MkRecord
  field
    aBool : Bool

  aBoolAsAFunction : Bool
  aBoolAsAFunction = aBool
open RecordTest public
{-# COMPILE AGDA2HS RecordTest newtype #-}

test : Bool
test = aBoolAsAFunction (MkRecord true)
{-# COMPILE AGDA2HS test #-}
```

And the output is:

```hs
module RecordTest where

newtype RecordTest = MkRecord{aBool :: Bool}

test :: Bool
test = aBoolAsAFunction
```

So the `(MkRecord true)` parameter gets omitted.

I've tried to take a look at `Record.hs`, but I think I don't understand it yet... So I would be grateful for any help:)

jespercockx:
Fixes #185

### DIFF
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -226,6 +226,7 @@ withFunctionLocals q k = do
       . map fst
       . filter (usableModality . getModality . snd) -- drop if it's an erased definition anyway
       . sortDefs <$> liftTCM curDefs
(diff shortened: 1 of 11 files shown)


viktorcsimma:
Nice; thank you very much:)

PR #209, opened 2023-10-22
Fixing Emacs mode

### LINKED ISSUE #208: Emacs mode not working since Agda 2.6.4 update
I have already changed the version numbers in the `.el` files from `2.6.3` to `2.6.4`, but there seems to be a bigger problem.

As far as I can see, the agda2hs binary does not support `--interaction` and `--disable-backend` used together. The error message is:
```
$ agda2hs --interaction --disable-backend input.agda
Error: Unrecognized option:
--disable-backend

Run 'agda2hs --help' for help on command line options.
```
The same message is what I get when trying to open a file in Emacs. Previously, the `--disable-backend` option is what was needed to use the `--interaction` option; now they don't seem to work together.

When only trying with the `--interaction` option:
```
$ agda2hs --interaction ErasureAndIrrelevance.agda
Error: Must not specify an input file (/home/csimma/Downloads/ErasureAndIrrelevance.agda) with --interaction
Run 'agda2hs --help' for help on command line options.
```
I don't know whether there has been a change in Agda 2.6.4 in the use of `--interaction`, or there is some other problem. Do you have any ideas? Thank you in advance!

viktorcsimma:
Fixes #208. The main function now passes a disabled backend to `runAgda` instead of no backends; maybe that was the problem previously.

### DIFF
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -56,7 +56,7 @@ backend = Backend'
   , options               = defaultOptions
   , commandLineFlags      =
       [ Option ['d'] ["disable-backend"] (NoArg disableOpt)
-        "Disable backend and fall back to vanilla Agda behaviour, without compilation (important for Emacs mode)."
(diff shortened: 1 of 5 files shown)


jespercockx:
Thank you!

PR #211, opened 2023-10-30
Fix for issue 210

### LINKED ISSUE #210: Phantom parameters of values in `where` clauses
**Update:** it seems to be wrong with functions, too. This test code:

```agda
open import Agda.Builtin.Nat

f1 : Nat -> Nat
f1 n = h1
  where
  g1 : Nat
  g1 = 3 + n
  h1 : Nat
  h1 = n + g1
{-# COMPILE AGDA2HS f1 #-}
```

gets compiled into this:

```hs
module PhantomParameters where

f1 :: Natural -> Natural
f1 n = h1
  where
    g1 :: Natural
    g1 = 3 + n
    h1 :: Natural
    h1 = n + g1 n
```

---

The following test code:

```agda
open import Agda.Builtin.Nat

record Test (a : Set) : Set₁ where
  field
    f : a -> a
open Test {{...}} public
{-# COMPILE AGDA2HS Test class #-}

instance
  testNat : Test Nat
  Test.f testNat n = h
    where
    g : Nat
    g = 3 + n
    h : Nat
    h = n + g
  {-# COMPILE AGDA2HS testNat #-}
```

gets compiled to this Haskell code:

```hs
module PhantomParameters where

class Test a where
    f :: a -> a

instance Test Natural where
    f n = h
      where
        g :: Natural
        g = 3 + n
        h :: Natural
        h = n + g n
```

So `g` gets an extra `n` parameter (probably because it depends on `n`). ~~Interestingly, this problem only occurs when writing instances; in normal functions, this is not a problem.~~

Commit 79fe184a did not contain the bug; so it has to be something that has happened since then.

jespercockx:
This fixes #210 by using the same way to track module parameters that should be dropped as Agda uses internally. It also removes support for functions inside record modules (as in the test case for #185) since these do not exist in Haskell in the first place.


jespercockx:
@viktorcsimma WDYT?

viktorcsimma:
It's exactly what I needed; thank you! I've tested it with my project and it works flawlessly.

I agree that if we have to choose between #210 and #185; #210 is more important. I used to depend on agda-stdlib where there sometimes are functions in record namespaces, but that's a quite rare use case.

PR #213, opened 2023-11-06
[ fix #212 ] Be more flexible with unboxed records

### LINKED ISSUE #212: Unboxed record not supported
The following Agda record:

```agda
record Sigma (@0 a : Set ℓ) (b : @0 a → Set ℓ′) : Set (ℓ ⊔ ℓ′) where
  constructor Pair
  field
    @0 proj1 : a
    proj2 : b proj1
```

(similar to `\exists` [here](https://github.com/agda/agda2hs/blob/9883f9648265c0967e70f9fd8085c0e6bc13f2b9/test/UnboxPragma.agda#L4), with the other field erased)
Gets compiled to the following Haskell code:

```haskell
data Sigma b = Pair {proj2 :: b}
```

And yet using the `COMPILE AGDA2HS Sigma unboxed` pragma errors out:

> An unboxed type must be a non-recursive record type with exactly
one non-erased field.

flupe:
Made the check for unboxed records more flexible. Now the unique non-erased field can be placed wherever.

### DIFF
--- a/src/Agda2Hs/Compile/Record.hs
+++ b/src/Agda2Hs/Compile/Record.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE NamedFieldPuns #-}
 module Agda2Hs.Compile.Record where
 
 import Control.Monad ( unless )
@@ -168,7 +169,11 @@ compileRecord target def = setCurrentRange (nameBindingSite $ qnameName $ defNam
       return $ Hs.DataDecl () don Nothing hd [conDecl] ds
 
 checkUnboxPragma :: Defn -> C ()
-checkUnboxPragma def@Record{ recFields = (f:fs) }
-  | keepArg f , all (not . keepArg) fs , not (recRecursive def) = return ()
-checkUnboxPragma _ = genericError $
-  "An unboxed type must be a non-recursive record type with exactly one non-erased field."
+checkUnboxPragma def
+  | Record{recFields} <- def
+  , length (filter keepArg recFields) == 1
+  , not (recRecursive def)
+  = return ()
+
+  | otherwise
+  = genericError "An unboxed type must be a non-recursive record type with exactly one non-erased field."
--- a/test/UnboxPragma.agda
+++ b/test/UnboxPragma.agda
@@ -10,6 +10,13 @@ open ∃ public
 
 {-# COMPILE AGDA2HS ∃ unboxed #-}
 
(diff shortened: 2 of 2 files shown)


src/Agda2Hs/Compile/Record.hs
```diff
@@ -168,7 +169,7 @@ compileRecord target def = setCurrentRange (nameBindingSite $ qnameName $ defNam
...
 
 checkUnboxPragma :: Defn -> C ()
-checkUnboxPragma def@Record{ recFields = (f:fs) }
-  | keepArg f , all (not . keepArg) fs , not (recRecursive def) = return ()
-checkUnboxPragma _ = genericError $
-  "An unboxed type must be a non-recursive record type with exactly one non-erased field."
+checkUnboxPragma def@Record{recFields} =
+  if length (filter keepArg recFields) == 1 && not (recRecursive def)
+    then return ()
+    else genericError "An unboxed type must be a non-recursive record type with exactly one non-erased field."
```
omelkonian, line 175: This will raise a `Non-exhaustive pattern matching` error now that you've removed the catch all clause.
The best option IMO is:
```hs
checkUnboxPragma def
  | Record{recFields} <- def
  , length (filter keepArg recFields) == 1
  , not (recRecursive def)
  = return ()
  | otherwise
  = genericError "An unboxed type must be a non-recursive record type with exactly one non-erased field."
```

test/UnboxPragma.agda
```diff
@@ -10,6 +10,13 @@ open ∃ public
 
 {-# COMPILE AGDA2HS ∃ unboxed #-}
 
+record Σ0 (A : Set) (P : @0 A → Set) : Set where
+  field
+    @0 el : A
+    pf : P el
+
+{-# COMPILE AGDA2HS Σ0 unboxed #-}
```
omelkonian, line 18: The associated golden file is missing.

omelkonian (changes requested):

src/Agda2Hs/Compile/Record.hs
flupe: Oops, indeed, let's keep the case!

test/UnboxPragma.agda
flupe: I expanded an existing test, and the new record shouldn't produce any output so the golden file should stay the same.

omelkonian (approved):
LGTM

PR #214, opened 2023-11-07
[fix #213] Properly deal with unboxed records, again

flupe:
Ok I should have tested my last PR more carefully indeed, sorry. *Using* an unboxed record in a type now also only keeps the type of the only non-erased field.


omelkonian (approved):
:+1:

PR #216, opened 2023-11-07
[fix #215] handle automatic projections from let pattern of tuples

### LINKED ISSUE #215: Tuple projection bug
The following Agda code:

```agda
pair : Int × Int
pair = (1 , 2)

{-# COMPILE AGDA2HS pair #-}

test : Int
test = let (x , y) = pair in x + y

{-# COMPILE AGDA2HS test #-}
```

Gets translated into the following:

```haskell
pair :: (Int, Int)
pair = (1, 2)

test :: Int
test = fst pair + fst (snd pair)
```

Somehow, the right component has an additional `fst` that makes it incorrect.

flupe:
The issue was coming from Agda inlining let bindings. This translation leaked our encoding for tuples.

The example from #215 now gets compiled properly, but:

- n-uples with n > 2 pattern-matched inside `let`s would still get compiled incorrectly because there is in fact no proper way to inline the `let` using projections (unless we also generate projections on the fly, like `\(_, _, x) -> x`).

  So `let (x, y, z) = e in z` would still compile to the incorrect `snd (snd e)`. We probably need a reliable way to prevent such use of pattern-matching on `let` for n-uples, doesn't have to be now.

  We could compute the length of projection chains of `Pair.fst` and `Pair.snd` in the terms we get, the only issue being that the resulting error would not point to the location of the `let`, only the location of the problematic variable.

- The projection `fst` and `snd` of `Pair` probably shouldn't get exported from `Haskell.Prim.Tuple` so as to enforce that the only way they pop up in terms is through inlining of `let` bindings.

Feedback welcome.

### DIFF
--- a/src/Agda2Hs/Compile/Term.hs
+++ b/src/Agda2Hs/Compile/Term.hs
@@ -5,7 +5,7 @@ import Control.Monad ( unless )
 import Control.Monad.Reader
 
 import Data.List ( isPrefixOf )
-import Data.Maybe ( fromMaybe, isJust )
+import Data.Maybe ( fromMaybe, isJust, listToMaybe )
 import qualified Data.Text as Text ( unpack )
 
 import qualified Language.Haskell.Exts as Hs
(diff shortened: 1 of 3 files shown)


jespercockx:
> n-uples with n > 2 pattern-matched inside lets would still get compiled incorrectly because there is in fact no proper way to inline the let using projections (unless we also generate projections on the fly, like \(_, _, x) -> x).

As we discussed, maybe we should just bite the bullet and define n-tuples for n =< 10 by hand.

PR #220, opened 2023-11-08
[fix #219] always qualify unopened imports

### LINKED ISSUE #219: Disappearing qualified import
Minimal example of a bug I encountered with qualified imports.

`A.agda`:
```agda
module A where

open import Haskell.Prelude

X : Set
X = ⊤

{-# COMPILE AGDA2HS X #-}
```

`B.agda`:
```agda
module B where

open import Haskell.Prelude

import A

X : Set
X = A.X

{-# COMPILE AGDA2HS X #-}
```

---

Compiling file `B.agda` yields:

```haskell
module B where

import A (X)

type X = X
```

`A` is not imported qualified somehow.

flupe:
The fix was easy enough, one branch for qualified imports (without a name) started returned `Nothing` after #202.

I changed a few lines that I found very hard to read, but can revert if it isn't to your taste. Also, can the lookup ever fail? I tried removing the `catchError` line locally and nothing happened on the test suite.

### DIFF
--- a/src/Agda2Hs/Compile/Name.hs
+++ b/src/Agda2Hs/Compile/Name.hs
@@ -150,10 +150,10 @@ compileQName f
         (C.Qual as C.QName{} : _) -> liftTCM $ do
           let qual = hsModuleName $ prettyShow as
           lookupModuleInCurrentModule as >>= \case
-            (x:_) | qual /= mod -> isDatatypeModule (amodName x) >>= \case
-              Just{} -> return $ QualifiedAs Nothing
-              Nothing -> return $ QualifiedAs $ Just qual
-            _ -> return Nothing
+            (x:_) | qual /= mod -> do
+              isDataMod <- isJust <$> isDatatypeModule (amodName x)
+              return $ QualifiedAs (if isDataMod then Nothing else Just qual)
+            _ -> return $ QualifiedAs Nothing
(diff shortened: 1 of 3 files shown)


src/Agda2Hs/Compile/Name.hs
```diff
@@ -150,10 +150,10 @@ compileQName f
...
           let qual = hsModuleName $ prettyShow as
           lookupModuleInCurrentModule as >>= \case
-            (x:_) | qual /= mod -> isDatatypeModule (amodName x) >>= \case
-              Just{} -> return $ QualifiedAs Nothing
-              Nothing -> return $ QualifiedAs $ Just qual
-            _ -> return Nothing
+            (x:_) | qual /= mod -> do
+              isDataMod <- isJust <$> isDatatypeModule (amodName x)
+              return $ QualifiedAs (if isDataMod then Nothing else Just qual)
+            _ -> return $ QualifiedAs Nothing
```
omelkonian, line 156: Indeed that's much clearer / easier to read.

omelkonian (approved):
LGTM

@jespercockx can you please confirm the removal of `catchError` was indeed redundant?

jespercockx:
Hm, the `catchError` might be an artifact left over from an earlier version of the function, it's probably safe to remove it (assuming there's no test case that it triggers).

jespercockx (approved):

PR #229, opened 2023-11-14
allow erased module arguments

### LINKED ISSUE #222: Allow erased module parameters
Haskell doesn't support module parameters, but those make management of Agda code way easier. So long as they are erased, they should be allowed.

### LINKED ISSUE #223: Functions defined with an absurd pattern get no definition in Haskell
If a function with a single clause is defined with an absurd pattern, only the type signature gets generated in the corresponding Haskell code.

```agda
impossible : @0 ⊥ → a
impossible ()
{-# COMPILE impossible #-}
```

```haskell
impossible :: a
```

We probably should have a default clause that throws a runtime error.

flupe:
Closes #222.

Also fixes #223 by throwing a hard error when a function gets no compiled clauses.

We now only add scoped type variables for definitions in modules that are not top-level, but they still get quantified over the parameters of the top-level module. Could be convenient, could be annoying, remains to be seen which behavior is the one we want.

This PR only officially supports erased parameters, or explicit sort parameters. Some other may be allowed but we don't have special checks I think (yet).

### DIFF
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -100,19 +100,32 @@ compileFun, compileFun' :: Bool -> Definition -> C [Hs.Decl ()]
 compileFun withSig def@Defn{..} = withFunctionLocals defName $ compileFun' withSig def
 -- inherit existing (instantiated) locals
 compileFun' withSig def@(Defn {..}) = do
-  reportSDoc "agda2hs.compile" 6 $ "compiling function: " <+> prettyTCM defName
+  reportSDoc "agda2hs.compile" 6 $ "Compiling function: " <+> prettyTCM defName
   when withSig $ whenJustM (liftTCM $ isDatatypeModule $ qnameModule defName) $ \_ ->
     genericDocError =<< text "not supported by agda2hs: functions inside a record module"
   let keepClause = maybe False keepArg . clauseType
   withCurrentModule m $ do
-    ifM (endsInSort defType) (ensureNoLocals err >> compileTypeDef x def) $ do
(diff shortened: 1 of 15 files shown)


src/Agda2Hs/Compile/Function.hs
```diff
@@ -100,19 +100,32 @@ compileFun, compileFun' :: Bool -> Definition -> C [Hs.Decl ()]
...
+        let filtered = filter keepClause funClauses
+        weAreOnTop <- isJust <$> liftTCM  (currentModule >>= isTopLevelModule)
         pars <- getContextArgs
-        reportSDoc "agda2hs.compile" 10 $ "applying clauses to parameters: " <+> prettyTCM pars
-        let clauses = filter keepClause funClauses `apply` pars
+        -- We only instantiate the clauses to the current module parameters
+        -- if the current module isn't the toplevel module
+        unless weAreOnTop $
+          reportSDoc "agda2hs.compile.type" 6 $ "Applying module parameters to clauses: " <+> prettyTCM pars
+        let clauses = if weAreOnTop then filtered else filtered `apply` pars
```
flupe, line 122: Could someone confirm that this is sufficient logic? It would appear top-level parameters are never part of a function clauses.

flupe:
I fixed the bug with module application. The output Haskell overly qualifies module applications, but Agda's prettyprinter does the same, so let's worry about it later.

flupe:
forgot to stage my new test file, hence the build failure...

jespercockx:
@flupe could you please resolve the conflicts in this PR so it can be included in 1.2?

jespercockx (approved):

Issue #135, opened 2023-01-04
Coercion compiled to unsafeCoerce


flupe:
In a recent implementation in Haskell, I had to use `unsafeCoerce`,
although it would have been provably safe if not for the lack of dependent types.

Jesper and I think it would be nice to have such a primitive in agda2hs:

```agda
coerce : @0 A ≡ B -> A -> B
```

that compiles to `unsafeCoerce`.

PR #237, opened 2023-11-28
Add `coerce` primitive

jakobn-ai:
Use `unsafeCoerce` when there is an (erased) proof that types are identical.

Fixes: #135

Q: (where) should this be documented?


jespercockx:
Thanks! The documentation should go into https://agda.github.io/agda2hs/features.html I guess, but that page really needs to be split up at some point.

jespercockx:
This seems simple and straightforward so I'm slipping it into the release of 1.2.

PR #238, opened 2023-12-01
Remove code duplication of agda-mode

omelkonian:
This PR removes the duplication originating in #198.

As discussed in [# 7012](https://github.com/agda/agda/issues/7012) of the Agda issue tracker, there is no need to duplicate all the Emacs-related code from the `agda/agda` repo; instead one can simply change the binary Emacs uses (I myself do it automatically using direnv and some custom variable initialisation elisp code in my `.emacs`).

@viktorcsimma Does this make sense or am I missing an advantage to copy-pasting this code in our repo?
@jespercockx As the reviewer of the original commit, WDYT?


jespercockx:
I approve of this, I think I was indeed to quick to accept the original PR.

jespercockx (approved):

PR #252, opened 2023-12-11
Add a primitive `Let` that is compiled to `let` in Haskell

jespercockx:


flupe:
May I suggest having a cursed syntax:

```agda
infixr 5 Let
syntax Let a (λ x → b) = let[ x ← a ] b

t : Bool
t =
  let[ x ← True  ]
  let[ y ← False ]
  x || y

```

flupe:
Note: Haskell's let bindings are *always* recursive, so we should make sure that the name used in the `Let` does not appear in the expression being bound to prevent incorrect behavior.

PR #253, opened 2023-12-12
Bring docs agda version in sync with .cabal file

liesnikov:
Docs say you need 2.6.3, cabal file says you need 2.6.4.

https://github.com/agda/agda2hs/blob/b269164e15da03b74cf43b51c522f4f052b4af80/agda2hs.cabal#L49-L50

(the code above is from v1.2 tag, but github refuses to render it if it's not a permalink)


jespercockx:
Thanks!

PR #254, opened 2023-12-15
add inlining pragma

flupe:
Unboxed records lead to the definition of many functions (on the Agda side) that lift functionality to unboxed records by wrapping and unwrapping accordingly. However, once compiled this overhead disappears and the resulting output could be inlined.

This PR introduces the `{-# COMPILE AGDA2HS name inline #-}` pragma that tells agda2hs to do the inlining automatically. Underapplied inline functions will be eta-expanded so that the substitution still happens.

---

```agda
record Wrap (a : Set) : Set where
  constructor Wrapped
  field
    unwrap : a
open Wrap public
{-# COMPILE AGDA2HS Wrap unboxed #-}

mapWrap : (f : a → b) → Wrap a → Wrap b
mapWrap f (Wrapped x) = Wrapped (f x)
{-# COMPILE AGDA2HS mapWrap inline #-}

mapWrap2 : (f : a → b → c) → Wrap a → Wrap b → Wrap c
mapWrap2 f (Wrapped x) (Wrapped y) = Wrapped (f x y)
{-# COMPILE AGDA2HS mapWrap2 inline #-}

test1 : Wrap Int → Wrap Int
test1 x = mapWrap (1 +_) x
{-# COMPILE AGDA2HS test1 #-}

test2 : Wrap Int → Wrap Int → Wrap Int
test2 x y = mapWrap2 _+_ x y
{-# COMPILE AGDA2HS test2 #-}

-- partial application of inline function
test3 : Wrap Int → Wrap Int → Wrap Int
test3 x = mapWrap2 _+_ x
{-# COMPILE AGDA2HS test3 #-}
```
compiles to:
```haskell
test1 :: Int -> Int
test1 x = 1 + x

test2 :: Int -> Int -> Int
test2 x y = x + y

test3 :: Int -> Int -> Int
test3 x = \ y -> x + y
```

---

Right now the names of the inserted lambdas come from the names given to the patterns in the inlined function. I suspect these could shadow names outside of where the inlined function is being used. Suggestions welcomed for making it more robust.

### DIFF
--- a/test/golden/Inline2.err
+++ b/test/golden/Inline2.err
@@ -1,2 +1,2 @@
 test/Fail/Inline2.agda:5,1-6
-Cannot make function tail' inlinable. Inline functions can only use variable patterns, dot patterns, or transparent record constructor patterns.
+Cannot make function tail' inlinable. Inline functions can only use variable patterns or transparent record constructor patterns.
--- a/test/golden/Inline2.err
+++ b/test/golden/Inline2.err
@@ -1,0 +1,2 @@
+test/Fail/Inline2.agda:5,1-6
+Cannot make function tail' inlinable. Inline functions can only use variable patterns, dot patterns, or transparent record constructor patterns.
--- a/src/Agda2Hs/Compile.hs
+++ b/src/Agda2Hs/Compile.hs
@@ -9,6 +9,7 @@ import qualified Data.Map as M
 import Agda.Compiler.Backend
 import Agda.Syntax.TopLevelModuleName ( TopLevelModuleName )
 import Agda.TypeChecking.Pretty
+import Agda.TypeChecking.Monad.Signature ( isInlineFun )
 import Agda.Utils.Null
 import Agda.Utils.Monad ( whenM )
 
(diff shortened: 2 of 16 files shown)


flupe:
Noticed it can loop on some examples, need to fix it.

jespercockx:
@flupe what's the status on this?

flupe:
Getting back to this now.

flupe:
Somehow #247 adding `test/golden/Haskell/...` made the CI unreliable, because the test themselves don't produce this file.
I thought we didn't cache test files but I removed the file anyway since this PR makes it unnecessary.

---

So the PR is ready. Instead of defining our own routine for inlining, we currently rely on Agda's `unfoldDefinitionStep` reduce function. This means we're overly conservative currently, as we forbid pattern-matching on anything other than unboxed records constructors.

It should be *fine* to allow pattern-matching on erased arguments (say, an equally type), but `unfoldDefinitionStep` would no longer unfold the function and we would have to implement our own matching routine. This is left for a next PR.

test/golden/Inline2.err
```diff
@@ -0,0 +1,2 @@
+test/Fail/Inline2.agda:5,1-6
+Cannot make function tail' inlinable. Inline functions can only use variable patterns, dot patterns, or transparent record constructor patterns.
```
jespercockx, line 2: Very minor comment: you cannot have dot patterns without having at least one constructor pattern with a constructor of an indexed datatype. So the mention of dot patterns here is redundant.

jespercockx:
Thanks!

PR #258, opened 2023-12-23
Adding support for explicit type signatures

viktorcsimma:
This is accomplished by the `the` function in Haskell.Prim. The first argument is the type, the second one is the value itself. For example, `the Nat 5` gets compiled to `5 :: Natural`.

Let me know what you think.


src/Agda2Hs/Compile/Term.hs
```diff
@@ -68,7 +73,18 @@ ifThenElse _ es = compileElims es >>= \case
...
   -- partially applied
-  _ -> genericError $ "if_then_else must be fully applied"
+  _ -> genericError $ "if_then_else_ must be fully applied"
+
+expTypeSig :: QName -> Elims -> C (Hs.Exp ())
+expTypeSig _ es = do
+  let args = fromMaybe __IMPOSSIBLE__ $ allApplyElims es
+  case args of
+    _ : typArg : expArg : args' -> do         -- the first one is the level; we throw that away
+      exp <- fromMaybe __IMPOSSIBLE__ <$> compileArg expArg -- this throws an error if it was Nothing
```
jespercockx, line 83: You're better off just doing `compileTerm . unArg`, which would avoid the call to `__IMPOSSIBLE__`

test/The.agda
```diff
@@ -0,0 +1,8 @@
+module The where
```
jespercockx, line 1: Maybe a name such as `TypeSignature` for the module would be more helpful?

jespercockx (approved):
Thanks for the PR! I have two small remarks but otherwise this seems like a really good feature to have. Since I've merged some other PRs in the meantime, could you also please merge/rebase on the current master?

viktorcsimma:
Done:)

jespercockx (approved):
Thank you!

PR #260, opened 2023-12-24
[ fix #257 ] fix recursive record check

### LINKED ISSUE #257: Unboxed record hard error
The following code produces a hard error (seemingly thrown by Agda, but most likely comes from agda2hs):

```agda
open import Haskell.Prelude

record BinInt : Set where
  field
    int : Integer
{-# COMPILE AGDA2HS BinInt unboxed #-}
```

```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Monad/Base.hs:2964:18 in Agda-2.6.4.1-d94eea85a1ca96c3755bdb85f3292bdd72fd5471203eca704b0e7066d091a850:Agda.TypeChecking.Monad.Base
```

If I remove the `unboxed`, no error. But I need the `unboxed`!

flupe:
#257 was caused by `recRecursive` being incorrectly implemented upstream.
Added a custom check for now, the old one could be put back once/if agda/agda#7042 is merged and the next version is released.

### DIFF
--- a/src/Agda2Hs/Compile/Record.hs
+++ b/src/Agda2Hs/Compile/Record.hs
@@ -170,9 +170,12 @@ compileRecord target def = do
 
 checkUnboxPragma :: Defn -> C ()
 checkUnboxPragma def
-  | Record{recFields} <- def
+  | Record{recFields, recMutual} <- def
   , length (filter keepArg recFields) == 1
-  , not (recRecursive def)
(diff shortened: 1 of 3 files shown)


flupe:
It would appear this could be reverted once Agda 2.7.0 is released, see: agda/agda#7149.

PR #261, opened 2023-12-28
Adding more documentation on Emacs mode

viktorcsimma:
Based on my own experiences, I tried to summarise how one can most easily install the new Emacs mode, with some instructions.

### DIFF
--- a/docs/source/features.md
+++ b/docs/source/features.md
@@ -1076,7 +1076,31 @@ If `implicit` is `false`, Prelude gets imported explicitly, and only those ident
 Since there is a full Agda typechecker in the `agda2hs` binary,
 you can readily use the Emacs mode of a pre-existing Agda installation (make sure it is the same version as the one `agda2hs` depends on) by changing the binary that `agda-mode` uses (Lisp variable `agda2-program-name`) to (the path to) `agda2hs` binary.
 
-With `C-c C-x C-c`, you will now be able call the `agda2hs` backend from inside Emacs; all the other built-in backends still remain available.
+- First, check the version of the Agda typechecker in the `agda2hs` binary by calling `agda2hs --version`.
+- Install this version of Agda. Perhaps the easiest way is to `git clone https://github.com/agda/agda`, checking out the tag corresponding to the correct version (e.g. `v2.6.4.1`; here `git fetch --tags` helps) and calling `cabal new-install`.
+- Run `agda-mode setup` and in Emacs, set the `agda2-program-name` variable to `agda2hs` (use `M-x customize-variable`). Alternatively, you can copy this into `~/.emacs`:
+
+```el
+(setenv "PATH" (concat (getenv "PATH") ":~/.cabal/bin"))
+  (setq exec-path (append exec-path '("~/.cabal/bin")))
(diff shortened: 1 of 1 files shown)


docs/source/features.md
```diff
@@ -1076,7 +1076,31 @@ If `implicit` is `false`, Prelude gets imported explicitly, and only those ident
...
+ ;; Your init file should contain only one such instance.
+ ;; If there is more than one, they won't work right.
+ '(agda2-program-name "agda2hs"))
+(custom-set-faces
+ ;; custom-set-faces was added by Custom.
+ ;; If you edit it by hand, you could mess it up, so be careful.
+ ;; Your init file should contain only one such instance.
+ ;; If there is more than one, they won't work right.
+ )
+```
```
omelkonian, line 1101: Since Emacs configurations drastically differ between users (e.g. I use `direnv` to set project-specific executables), I wouldn't put this here and let the users configure as they see fit.

docs/source/features.md
```diff
@@ -1076,7 +1076,31 @@ If `implicit` is `false`, Prelude gets imported explicitly, and only those ident
 Since there is a full Agda typechecker in the `agda2hs` binary,
 you can readily use the Emacs mode of a pre-existing Agda installation (make sure it is the same version as the one `agda2hs` depends on) by changing the binary that `agda-mode` uses (Lisp variable `agda2-program-name`) to (the path to) `agda2hs` binary.
 
-With `C-c C-x C-c`, you will now be able call the `agda2hs` backend from inside Emacs; all the other built-in backends still remain available.
+- First, check the version of the Agda typechecker in the `agda2hs` binary by calling `agda2hs --version`.
+- Install this version of Agda. Perhaps the easiest way is to `git clone https://github.com/agda/agda`, checking out the tag corresponding to the correct version (e.g. `v2.6.4.1`; here `git fetch --tags` helps) and calling `cabal new-install`.
+- Run `agda-mode setup` and in Emacs, set the `agda2-program-name` variable to `agda2hs` (use `M-x customize-variable`). Alternatively, you can copy this into `~/.emacs`:
```
omelkonian, line 1081: This repeats the same information with the previous paragraph and we shouldn't include Agda installation instructions ourselves.

omelkonian (commented):
IMO this is addition to the documentation is unnecessary.

One possible improvement you could retain from this PR is some explanatory additions to the existing paragraph (above the ones you added), e.g. to check agda2hs's version using `agda2hs --version` and customizing the Emacs executable with `M-x customize-variable`, but nothing else.

viktorcsimma:
What about this? I've left the version extraction and the setting of the variable.

omelkonian:
Still a lot of duplication with the previous paragraph; can you remove all the repeated content there? 

Something like:
```
Since there is a full Agda typechecker in the `agda2hs` binary,
you can readily use the Emacs mode of a pre-existing Agda installation by changing the binary that `agda-mode` uses to the path to `agda2hs` binary.
```

viktorcsimma:
Done:)

Also, the third point was split into two to bring attention to the two different steps.

PR #263, opened 2024-01-15
Add built-in support for Delay monad

jespercockx:
Currently there is no good way to define potentially-looping functions with `agda2hs`. We have the `loop` function in `Haskell.Extra.Loop` but it still requires a proof of termination if you want to use it on the Agda side. This PR adds support for the Delay monad to agda2hs, in a way that is completely transparent in the resulting Haskell code.

### DIFF
--- a/lib/Haskell/Extra/Delay.agda
+++ b/lib/Haskell/Extra/Delay.agda
@@ -1,0 +1,39 @@
+{-# OPTIONS --sized-types #-}
+
+module Haskell.Extra.Delay where
+
+open import Agda.Builtin.Size public
+
+open import Haskell.Prelude
+open import Haskell.Prim.Thunk
+open import Haskell.Extra.Refinement
+
+private variable
+  x y z : a
+  @0 i : Size
+
+data Delay (a : Set) (@0 i : Size) : Set where
+  now : a → Delay a i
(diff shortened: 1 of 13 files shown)


jespercockx:
Hm, I just realized this is not sound, since it is possible to distinguish between `now` and `later` in Agda. For example, one can write the following function:

```agda
bad : Delay a ∞ → Bool
bad (now x) = True
bad (later x) = False
```
which would get compiled to
```haskell
bad :: a -> Bool
bad x = True
bad x = False
```
which is, well, bad.

If we want to have actual support for `delay`, we should enforce somehow that the result is the same in both cases, or simply forbid pattern matching on `Delay` altogether.

jespercockx:
Perhaps we could allow a single function `runDelay` defined like this:
```agda
runDelay : {@0 x : a} (y : Delay a ∞) → @0 HasResult x y → a
runDelay (now x) now = x
runDelay (later y) (later p) = runDelay (y .force) p
```
plus possibly a proof that the result of `runDelay` is indeed equal to `x`.

jespercockx:
@flupe this is ready now, could you do a quick review?

flupe:
lgtm

flupe (approved):

PR #270, opened 2024-01-23
type-directed compilation

### LINKED ISSUE #217: Allow non-erased hidden arguments (but make them explicit)
Currently, agda2hs doesn't allow functions to have _non-erased_ hidden arguments, unless they are `Set` or `Level`. I would like to propose we allow other kind of hidden arguments: we simply make them explicit on the Haskell side.

While this doesn't fit the philosophy of only making Haskell patterns available in Agda, I think there is a good motivation for it when doing intrinsic verification: [tactic arguments](https://agda.readthedocs.io/en/latest/language/implicit-arguments.html#tactic-arguments).

We can use such tactics on the Agda side to lighten the syntax, even though we care about the solutions on the Haskell side.

---

The example I have in mind is Jesper's [syntax for agda-core](https://github.com/jespercockx/agda-core/blob/main/Syntax.agda#L37):

```agda
data Term α where
  var    : (@0 x : Name) → {@(tactic auto) x∈α : x ∈ α} → Term α
```

Such a definition permits using names on the Agda side and letting Agda figure out indices in the global scope using a tactic. However, we want names to be erased and the index to be preserved.

flupe:
A big refactoring as a first step to properly handle module parameters.

The most important change is the following: the compilation of terms if now completely type-directed. That is, we no longer rely on `ArgInfo` of `Elims` to know whether they have to be preserved, but rather their `Dom Type` only.

What this means, concretely:
- Explicit usable erasable arguments (Say, `(a : Set)`) are dropped.
- Implicit usable non-erasable arguments (Say, `{x : Nat}`) are made explicit.  
  (This means we now officially support tactic arguments that are made explicit on the Haskell side.)

As a treat, we now properly erase types of sort `Prop`.

Closes #217.

---

TODO list before opening the PR:

- [x] Add back special class functions.
- [x] Fix compilation of co-patterns.
- [x] Fix compilation of patterns on projection-like functions.
- [x] Add back compilation of inline functions.
- [x] Ensure unfolding is back.
- [x] Drop module parameters again.

### DIFF
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -52,12 +52,23 @@ import Agda.TypeChecking.Datatypes (isDataOrRecord)
 -- | Compilation rules for specific constructors in patterns.
 isSpecialCon :: QName -> Maybe (Type -> NAPs -> C (Hs.Pat ()))
 isSpecialCon qn = case prettyShow qn of
+  s | s `elem` badConstructors     -> Just itsBad
   "Haskell.Prim.Tuple._,_"         -> Just tuplePat
   "Haskell.Prim.Tuple._×_×_._,_,_" -> Just tuplePat
   "Haskell.Extra.Erase.Erased"     -> Just tuplePat
   "Agda.Builtin.Int.Int.pos"       -> Just posIntPat
   "Agda.Builtin.Int.Int.negsuc"    -> Just negSucIntPat
   _                                -> Nothing
+  where
+    badConstructors =
+      [ "Agda.Builtin.Nat.Nat.zero"
+      , "Agda.Builtin.Nat.Nat.suc"
+      , "Haskell.Extra.Delay.Delay.now"
+      , "Haskell.Extra.Delay.Delay.later"
+      ]
+
+    itsBad :: Type -> NAPs -> C (Hs.Pat ())
+    itsBad _ _ = genericDocError =<< "constructor `" <> prettyTCM qn <> "` not supported in patterns"
 
 -- | Translate list of patterns into a Haskell n-uple pattern.
 tuplePat :: Type -> NAPs -> C (Hs.Pat ())
@@ -92,23 +103,6 @@ compileLitNatPat = \case
     , [p] <- ps -> (1+) <$> compileLitNatPat (namedArg p)
   p -> genericDocError =<< "not a literal natural number pattern:" <?> prettyTCM p
 
-  {-
-  case prettyShow qn of
-  s | s `elem` badConstructors     -> Just $ \ _ -> genericDocError =<<
-    "constructor `" <> prettyTCM qn <> "` not supported in patterns"
-  _                                -> Nothing
-  where
-    badConstructors =
-      [ "Agda.Builtin.Nat.Nat.zero"
-      , "Agda.Builtin.Nat.Nat.suc"
-      , "Haskell.Extra.Delay.Delay.now"
-      , "Haskell.Extra.Delay.Delay.later"
-      ]
-
-
-
-
--}
 
 compileFun, compileFun'
   :: Bool -- ^ Whether the type signature shuuld also be generated
@@ -218,10 +212,21 @@ compileClause' curModule x ty c@Clause{..} = do
 compilePats :: Type -> NAPs -> C [Hs.Pat ()]
 compilePats _ [] = pure []
 compilePats ty ((namedArg -> ProjP po pn):ps) = do
+  reportSDoc "agda2hs.compile" 6 $ "compiling copattern: " <+> text (prettyShow pn)
+  -- NOTE: should be fine for unboxed records
+  unlessM (asks copatternsEnabled `or2M` (isJust <$> isUnboxProjection pn)) $
+    genericDocError =<< "not supported in Haskell: copatterns"
   Just (unEl -> Pi a b) <- getDefType pn ty -- ????
   compilePats (absBody b) ps
 
--- TODO: use shouldBePi or mustBePi
+-- -- copatterns patterns
+-- compilePat ty (ProjP _ q) = do
+--   reportSDoc "agda2hs.compile" 6 $ "compiling copattern: " <+> text (prettyShow q)
+--   unlessM (asks copatternsEnabled) $
+--     genericDocError =<< "not supported in Haskell: copatterns"
+--   let x = hsName $ prettyShow q
+--   return $ Hs.PVar () x
+
 compilePats ty ((namedArg -> pat):ps) = do
   (a, b) <- mustBePi ty
   reportSDoc "agda2hs.compile.pattern" 5 $ text "Compiling pattern:" <+> prettyTCM pat
@@ -261,13 +266,6 @@ compilePat ty (ConP ch i ps) = do
 -- literal patterns
 compilePat ty (LitP _ l) = compileLitPat l
 
--- copatterns patterns
-compilePat ty (ProjP _ q) = do
-  reportSDoc "agda2hs.compile" 6 $ "compiling copattern: " <+> text (prettyShow q)
-  unlessM (asks copatternsEnabled) $
-    genericDocError =<< "not supported in Haskell: copatterns"
-  let x = hsName $ prettyShow q
-  return $ Hs.PVar () x
 
 -- nothing else is supported
 compilePat _ p = genericDocError =<< "bad pattern:" <?> prettyTCM p
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -45,7 +45,7 @@ import Agda2Hs.HsUtils
 import Agda.TypeChecking.Datatypes (isDataOrRecord)
 
 
--- | Pattern compilation rules for specific constructors.
+-- | Compilation rules for specific constructors in patterns.
 isSpecialCon :: QName -> Maybe (NAPs -> C (Hs.Pat ()))
 isSpecialCon qn = case prettyShow qn of
   "Haskell.Prim.Tuple._,_"         -> Just tuplePat
@@ -125,14 +125,15 @@ compileFun' withSig def@Defn{..} = do
 
   withCurrentModule m $ do
     ifM (endsInSort defType)
-
         -- if the function type ends in Sort, it's a type alias!
         (ensureNoLocals err >> compileTypeDef x def) 
-
         -- otherwise, we have to compile clauses.
         $ do
+
       when withSig $ checkValidFunName x
+
       compileTopLevelType withSig defType $ \ty -> do
+
         let filtered = filter keepClause funClauses
         weAreOnTop <- isJust <$> liftTCM  (currentModule >>= isTopLevelModule)
         pars <- getContextArgs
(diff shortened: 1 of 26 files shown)


flupe:
All the tests pass again, finally.

Before reviewing this PR, there's still one caveat to fix: currently every test has the flag `--no-projection-like` set, because projection-like are annoying to manipulate.

The following things are likely to break once I remove the flag:
- [ ] Compilation of functions, as projection-like only store patterns that come after the record argument.
- [ ] Compilation of `Def`s, because some defs will be compiled in `compileProj`.
- [ ] `compileSpined`?

src/Agda2Hs/Compile/Type.hs
```diff
@@ -210,16 +209,30 @@ compileInlineType f args = do
...
-            (return DomDropped)
-            (genericDocError =<< do text "Implicit type argument not supported: " <+> prettyTCM x)
-  | otherwise    = return DomDropped
+      _          -> ifM (canErase $ unDom a) (return DomDropped)
+                        (uncurry DomType <$> compileTypeStrictness (unEl $ unDom a))
+  | otherwise = return DomDropped
+
+data DomOutput = DOInstance | DODropped | DOKept
+
+compiledDom :: Dom Type -> C DomOutput
```
jespercockx, line 229: Why is this a separate function from `compileDom`? Seems like it's a potential source of subtle bugs if these two don't match.

src/Agda2Hs/Compile.hs
```diff
@@ -24,81 +25,67 @@ import Agda2Hs.Compile.Types
...
 import Agda2Hs.Pragma
 
--- Needs a list of rewrite rules too.
+
 initCompileEnv :: TopLevelModuleName -> SpecialRules -> CompileEnv
 initCompileEnv tlm rewrites = CompileEnv
-  { currModule = tlm
-  , minRecordName = Nothing
-  , locals = []
+  { currModule        = tlm
```
jespercockx, line 31: It would be a bit nicer to bundle all cosmetic changes such as this one in a separate PR, since it will make it easier to locate any bugs that might be introduced.

src/Agda2Hs/Compile/Utils.hs
```diff
@@ -46,6 +47,36 @@ import Agda2Hs.Compile.Types
...
+    Var i es     -> fromMaybe (tm, ty) <$> loop (Var i) es
+    Def f es     -> undefined 
+    Con ch ci es -> undefined 
+    _ -> pure (tm, ty)
+  where
+    loop :: (Elims -> Term) -> Elims -> C (Maybe (Term, Type))
+    loop h es =
+      case es of
+        []            -> pure Nothing
+        Proj o f : es -> undefined -- Just $ fromMaybe (Def f (Apply (defaultArg v) : es')) $ loop h (Proj o f : res) es'
```
jespercockx, line 76: This looks like it should still be fixed.

src/Agda2Hs/Compile/Utils.hs
```diff
@@ -46,6 +47,36 @@ import Agda2Hs.Compile.Types
...
+  ]
+
+-- | Convert the final 'Proj' projection elimination into a
+--   'Def' projection application, but typed.
+unSpine1' :: Type -> Term -> C (Term, Type)
+unSpine1' ty tm =
+  case tm of
+    -- Just (h, es) -> fromMaybe v $ loop h [] es
+    Var i es     -> fromMaybe (tm, ty) <$> loop (Var i) es
+    Def f es     -> undefined
```
jespercockx, line 68: This looks like it should still be fixed.

src/Agda2Hs/Compile/Utils.hs
```diff
@@ -287,3 +330,56 @@ checkSingleElement name fs s = unless (length fs == 1) $ genericDocError =<< do
 
 checkingVars :: C a -> C a
 checkingVars = local $ \e -> e { checkVar = True }
+
+checkFixityLevel :: QName -> FixityLevel -> C (Maybe Int)
```
jespercockx, line 334: Could this stuff related to fixities perhaps be moved to a separate PR?

src/Agda2Hs/Render.hs
```diff
@@ -75,11 +76,15 @@ moduleFileName opts name = do
   outDir <- compileDir
   return $ fromMaybe outDir (optOutDir opts) </> moduleNameToFileName name "hs"
 
+-- NOTE: what does this have to do with rendering?
```
jespercockx, line 79: Should probably be moved to the top-level Compile module

test/UnboxPragma.agda
```diff
@@ -22,7 +22,7 @@ sort1 xs = xs [ looksfine ]
 sort2 : List Int → ∃ (List Int) IsSorted
 sort2 xs .el = xs
 sort2 xs .pf = looksfine
-
+
```
jespercockx, line 25: Please fix whitespace

src/Agda2Hs/Compile/Function.hs
```diff
@@ -163,103 +165,122 @@ compileFun' withSig def@(Defn {..}) = do
...
+--             so we should drop the first parameters in the input type (using funProjection.projLams)
+-- TODO(flupe): handle copatterns (that don't expect a Pi type) (See Unbox:sort2)
+compilePats :: Type -> NAPs -> C [Hs.Pat ()]
+compilePats _ [] = pure []
+compilePats ty ((namedArg -> ProjP po pn):ps) = do
+  reportSDoc "agda2hs.compile" 6 $ "compiling copattern: " <+> text (prettyShow pn)
+  -- NOTE: should be fine for unboxed records
+  unlessM (asks copatternsEnabled `or2M` (isJust <$> isUnboxProjection pn)) $
+    genericDocError =<< "not supported in Haskell: copatterns"
+  Just (unEl -> Pi a b) <- getDefType pn ty -- ????
```
jespercockx, line 224: Shouldn't you use `mustBePi` here to account for possible reduction?

src/Agda2Hs/Compile/Term.hs
```diff
@@ -20,311 +21,446 @@ import Agda.Syntax.Internal
...
+
+compileTerm :: Type -> Term -> C (Hs.Exp ())
+compileTerm ty v = do
+
+  reportSDoc "agda2hs.compile" 7  $ text "compiling term:" <+> prettyTCM v
+  reportSDoc "agda2hs.compile" 27 $ text "compiling term:" <+> pure (P.pretty $ unSpine1 v)
+
+  case compileSpined ty v of
+    Just cont -> cont
+    Nothing   -> case unSpine1 v of
```
jespercockx, line 395: Is `unSpine1` here still necessary? All cases with a spine are already handled by `compileSpined`.

jespercockx (commented):
Great work! This is definitely a major improvement to the compilation process of `agda2hs`. Going through the changes also made me realize that we're really not dealing with a small project here anymore. I have a couple of remarks and I think the projection-like functions should still be fixed. But overall I think this should be merged sooner rather than later.

flupe:
Rebased on #277.

src/Agda2Hs/Compile/Type.hs
flupe: The reason was `compileDom` atually compiles all the types to Haskell, so if this was the one being used in the typed-compilation we would be constantly compiling types to Haskell, even though we only care about whether those things are kept or dropped.
If we had dependent types it would be easier to express at the type-level that `compileDom` is now a refinement of `compiledDom`, but I'll just implement the former with the latter.

src/Agda2Hs/Compile/Term.hs
```diff
@@ -20,313 +21,448 @@ import Agda.Syntax.Internal
...
+-- | Compile @the@ to an explicitly-annotated Haskell expression.
+expTypeSig :: DefCompileRule
+expTypeSig ty args@(_:typ:_:_) = do
+    annot    <- compileType typ
+    exp:args <- compileArgs ty args
+    pure (Hs.ExpTypeSig () exp annot `eApp` args)
+expTypeSig _ _ = genericError "`the` must be fully applied"
+
+
+-- should really be named compileVar, TODO: rename compileVar
```
omelkonian, line 119: I suggest `s/compileVar/compileVarName` and then `s/compileV/compileVar`.

omelkonian (commented):
Great work Lucas!

One thing that slightly worries me is that I would expect this refactoring to reduce code size, but it in fact increases it substantially (c.f. [+800/-500](https://github.com/flupe/agda2hs/compare/cosmetic...flupe:agda2hs:modules)). Factoring out new tests and utilities we did not need before, it seems the main culprit is `Compile.Term`; do you think the final polished version might be able to reduce LOC down a bit?

PS: Ping me when the projection-like optimization has been properly handled.

jakobn-ai:
Implicit arguments with instance functions are still a compiler error, is this expected? i.e.

```agda
test : {x : Nat} → Nat
test {x} = x
{-# COMPILE AGDA2HS test #-}
```

will work but

```agda
instance
  test : {x : Nat} → Nat
  test {x} = x
{-# COMPILE AGDA2HS test #-}
```

will give

```
Location of the error: __IMPOSSIBLE__, called at src/Agda2Hs/Compile/ClassInstance.hs:103:20 in main:Agda2Hs.Compile.ClassInstance
```

flupe:
Hmm, to be fair the compilation of class instances is the part I'm the least happy with currently (the current implementation and what I did to make it typed).

Though I fail to see when a *compiled* instance with hidden relevant arguments turning relevant would ever make sense. I will add a check that for instances specifically, no relevant arguments are ever hidden as I think this would do the trick. Thanks for the example!

If you ever encounter an example where this would be useful behavior, hit me up!

src/Agda2Hs/Compile/Term.hs
flupe: `compileVar` as renamed to `compileDBVar` in #277 . Renamed `compileV` to `compileVar`, lmk if that's enough.

flupe:
@omelkonian: I fixed the compilation of projection-like functions and addressed all of your comments!

You will see that one test still requires the `--no-projection-like` flag as compiling projection-like (or projections) directly applied to constructors in a typed setting kinda goes against bidirectional tc. Quoting @jespercockx:

> a projection-like function applied to a constructor violates bidirectionality so you cannot (easily) do a type-aware traversal of it

If you remove the flag, the code will still compile to correct Haskell, but the projection-like function `unbind` will get reduced. Not incorrect, but just not what the user wrote.

If someone is really motivated, perhaps using [projView](https://hackage.haskell.org/package/Agda-2.6.4.1/docs/Agda-TypeChecking-ProjectionLike.html#v:projView) could allow you to preserve the projection-like function application, but I think you'd still have to trigger inference at some point, which I don't want to do.

> One thing that slightly worries me is that I would expect this refactoring to reduce code size, but it in fact increases it substantially (c.f. [+800/-500](https://github.com/flupe/agda2hs/compare/cosmetic...flupe:agda2hs:modules)). Factoring out new tests and utilities we did not need before, it seems the main culprit is Compile.Term; do you think the final polished version might be able to reduce LOC down a bit?

Hm, going into this implementation I had the same assumption, but quickly realized some more work was unavoidable (e.g Defs and projections cannot really be treated uniformly anymore).
All in all, I don't think too much code is added, there is also a lot of whitespace introduced because I like my code to breathe.

Hopefully things should be more easily maintainable too :)

----

>  @jakobn-ai: came across this experimenting with adding runtime checks for e.g. erased preconditions. A deciding version of a precondition would ideally be written as instance so it can be found in instance search: [...]

But the instance itself shouldn't be compiled to agda2hs right? Or using some custom logic (that you'd have to implement yourself).

I decided against adding the check I mentioned above, since the class compilation routine already doesn't enforce that, say, we don't pattern-match on parameters of instances, and they are silently dropped til we reach a (class) projection ([See here](https://github.com/agda/agda2hs/blob/4e3b9a73c87c03140f95d817fee29618333d1dcb/src/Agda2Hs/Compile/ClassInstance.hs#L138)).

So something should be done alright, just not in this PR!

omelkonian:
> If you remove the flag, the code will still compile to correct Haskell, but the projection-like function `unbind` will get reduced. Not incorrect, but just not what the user wrote.

Could you also add these unit tests, i.e. the ones without the flag where the generated code is further away from what the user wrote?

Then, a subsequent PR that improves on that (e.g. using `projView`) could improve those tests.

flupe:
Done! (See `test/ProjLike.agda`)

Issue #54, opened 2021-04-21
Bug: Miscompilation of operator precedence


JonathanBrouwer:
When compiling this code: 
```agda
module Testing where

open import Haskell.Prelude

infix 8 _^_ -- Higher priority than +, the usual precedence
_^_ : Nat -> Nat -> Nat
_^_ b zero = 1
_^_ b (suc e) = b * (b ^ e) 
{-# COMPILE AGDA2HS _^_ #-}

x = 2 ^ (2 + 3)
{-# COMPILE AGDA2HS x #-}
```
The following haskell is generated
```haskell
module Testing where

import Numeric.Natural (Natural)

(^) :: Natural -> Natural -> Natural
b ^ zero = 1
b ^ suc e = b * b ^ e

x :: Natural
x = 2 ^ 2 + 3
```

Note that there is nothing funky going on with the precedence, it's exactly as would be expected, yet it generates wrong code.
In fact, I can never get agda2hs to generate parentheses in infix sequences, no matter the parentheses used in the agda code

JonathanBrouwer:
You don't even need a custom operator definition, even `10 - (2 + 3)` does not compile correctly

omelkonian:
Yep, this is a feature we do not currently support -- see the un-checked item in the [README](https://github.com/agda/agda2hs#future-work).

Closing as this is a duplicate of #41.

jespercockx:
Even without using explicit operator precedences, shouldn't at least an expression such as `10 - (2 + 3)` (with all parenthesis included) be compiled to an equivalent expression in Haskell?

UlfNorell:
Yes, handling the default fixities is a slightly different issue than handling user-given ones.

PR #276, opened 2024-02-09
[ fix #273 ] Add more parenthesis in arguments of operators and sections

### LINKED ISSUE #273: `snd $` compiles incorrectly
There seems to be an issue when compiling the function application operator `$` together with `snd`:

The file

```agda
module Test where

open import Haskell.Prelude

test : Int × Int → Int
test = λ x → snd $ x

{-# COMPILE AGDA2HS test #-}
```

is compiled to 

```haskell
module Test where

test :: (Int, Int) -> Int
test = (\ r -> snd r $)
```

which is **not** valid Haskell.

Apparently, something went wrong with the fixity of `$` and the compilation of `snd` (Agda) to `\ r -> snd r` (Haskell). It appears to me that a parentheses are missing, a good compilation output would be

```haskell
test = ((\ r -> snd r) $)
```

Agda2hs version: `github:agda/agda2hs?ref=v1.2`

jespercockx:
Ideally, our pretty-printing library should insert parenthesis where needed. However, hs-src-exts does not insert adequate parenthesis for infix operators so we need to insert some by hand (see issues #54 and #273).

This also disables the printing of `_- u` as a right section `(- u)` in Haskell, as that would be interpreted as a negation instead.

### DIFF
--- a/src/Agda2Hs/Compile/Term.hs
+++ b/src/Agda2Hs/Compile/Term.hs
@@ -282,14 +282,14 @@ compileTerm v = do
       -- System-inserted lambda, no need to preserve the name.
       underAbstraction_ b $ \ body -> do
         x <- showTCM (Var 0 [])
-        let hsx = hsVar x
         body <- compileTerm body
         return $ case body of
-          Hs.InfixApp _ a op b
-            | a == hsx
-            , pp op /= "-"             -- Jesper: no right section for minus, as Haskell parses this as negation!
-            -> Hs.RightSection () op b -- System-inserted visible lambdas can only come from sections
-          _            -> hsLambda x body         -- so we know x is not free in b.
+          Hs.InfixApp _ a op b | a == hsVar x ->
+            if pp op == "-" then -- Jesper: no right section for minus, as Haskell parses this as negation!
+              Hs.LeftSection () b (Hs.QConOp () $ Hs.UnQual () $ hsName "subtract")
+            else                       -- System-inserted visible lambdas can only come from sections
+              Hs.RightSection () op b  -- so we know x is not free in b.
+          _            -> hsLambda x body         
     Lam v b ->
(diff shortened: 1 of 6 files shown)


src/Agda2Hs/Compile/Term.hs
```diff
@@ -286,7 +286,9 @@ compileTerm v = do
         body <- compileTerm body
         return $ case body of
           Hs.InfixApp _ a op b
-            | a == hsx -> Hs.RightSection () op b -- System-inserted visible lambdas can only come from sections
+            | a == hsx
+            , pp op /= "-"             -- Jesper: no right section for minus, as Haskell parses this as negation!
+            -> Hs.RightSection () op b -- System-inserted visible lambdas can only come from sections
```
omelkonian, line 291: Introducing `\ section -> section - ...` is quite ugly; why not use a left section on [subtract = flip (-)](https://hackage.haskell.org/package/base-4.7.0.2/docs/Prelude.html#v:subtract) instead?
```hs
            | a == hsVar x
            -> if pp op == "-" then
                   -- no right section for minus, as Haskell parses this as negation!
                   Hs.LeftSection () (QConOp $ unQual "subtract") b 
                else
                   -- System-inserted visible lambdas can only come from sections
                   Hs.RightSection () op b
```

Minor comments:
1. No need for the intermediate variable `hsx`, just inline here
2. No need to mention the name of the person writing the comment :P

jespercockx: For this PR I wasn't really looking at whether the output is *pretty*, just whether it is *correct*. But I agree that your version is prettier.

The variable `hsx` was already there, I didn't want to touch it since it's unrelated to this PR.

Writing my name in comments is a habit I got from working on Agda itself where it seems to be standard practice. I personally quite like these little bits of personality in the codebase :)

PR #282, opened 2024-02-16
Add some general introduction to the index page of the documentation

jespercockx:
People were complaining that `agda2hs` is not a general-purpose Haskell backend and is not compatible with stdlib, so hopefully this will make it clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,40 +2,14 @@
 
 # agda2hs
 
-Agda2hs is a tool for producing verified and readable Haskell code by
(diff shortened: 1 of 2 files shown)


omelkonian (commented):
Can you remove the duplicated context in the README?

jespercockx:
I removed the example from the README to reduce the duplication. The rest of the information feels like it is useful to have in the README and is anyway not that long, so I'm inclined to keep it. If you like to make further changes to the README, please make a new PR for that.

Issue #218, opened 2023-11-07
Allow instance arguments for non-typeclass records


flupe:
Similarly to #217, instance arguments and search is often used when doing intrinsinc verification. However, the data being searched may be runtime-relevant (i.e non-erased) and not a typeclass instance.

I suggest we let `agda2hs` keep track of whether a record has been compiled to a class or a Haskell data, and allow instance arguments of the latter, as long as we turn them explicit.

One such example is `Rezz` for [resurecting erased values](https://github.com/jespercockx/agda-core/blob/main/Utils.agda#L294):

```agda
record Rezz (a : Set) (@0 x : a) : Set where
  constructor Rezzed
  field
    rezzed : a
    @0 isRezzed : y ≡ x

open Rezz public

{-# COMPILE AGDA2HS Rezz unboxed #-}
```

It gets compiled unboxed (i.e `Rezz a x` gets compiled to `a`). Yet, instance search can be used on the Agda side to avoid having to construct evidence manually.

jespercockx:
Same comment as for #217, I think this is a sensible thing to add support for.

Issue #251, opened 2023-12-08
Check that types of non-erased instance arguments are type classes


jespercockx:
Currently we do not check that the type of an instance argument is actually compiled to a Haskell `Constraint`, causing `agda2hs` to generate ill-formed Haskell code. Here's an example:
```agda
open import Haskell.Prelude

instance
  favoriteNumber : Int
  favoriteNumber = 42
{-# INLINE favoriteNumber #-}

test : {{Int}} → Int
test {{x}} = x
{-# COMPILE AGDA2HS test #-}
```
Currently `agda2hs` happily compiles this to the following nonsense:
```haskell
test :: Int => Int
test = x
```
We should either disallow this, or else detect that the type has kind `*` and compile it to a regular (explicit) argument in Haskell.

jespercockx:
Related: #217 and #218

PR #284, opened 2024-02-21
Only compile Agda instances to a Haskell instance if they target a Haskell class

jespercockx:
This is a little gift for @jakobn-ai , who asked in https://github.com/agda/agda2hs/pull/270#issuecomment-1948366775 if we could have instances on the Agda side that compile to regular functions in Haskell. I thought this should be easy since we can just check whether a given instance actually targets a type that is compiled to a Haskell class, and it turns out I was right :)

### DIFF
--- a/src/Agda2Hs/Compile.hs
+++ b/src/Agda2Hs/Compile.hs
@@ -25,7 +25,7 @@ import Agda2Hs.Compile.Function ( compileFun, checkTransparentPragma, checkInlin
 import Agda2Hs.Compile.Postulate ( compilePostulate )
 import Agda2Hs.Compile.Record ( compileRecord, checkUnboxPragma )
 import Agda2Hs.Compile.Types
-import Agda2Hs.Compile.Utils ( setCurrentRangeQ, tellExtension, primModules )
+import Agda2Hs.Compile.Utils ( setCurrentRangeQ, tellExtension, primModules, isClassName )
 import Agda2Hs.Pragma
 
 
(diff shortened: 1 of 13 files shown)


jespercockx:
The second commit implements the same logic but for instance *arguments*: if the type of an argument is not compiled to a class in Haskell, then we compile it to a regular argument instead.

src/Agda2Hs/Compile.hs
```diff
@@ -80,23 +80,27 @@ compile opts tlm _ def =
...
-        (DefaultPragma ds    , _      , Datatype{}) -> compileData False ds def
-        (DerivePragma s      , Just _ , _         ) -> pure <$> compileInstance (ToDerivation s) def
-        (DefaultPragma _     , Just _ , Axiom{}   ) -> pure <$> compileInstance (ToDerivation Nothing) def
-        (DefaultPragma _     , Just _ , _         ) -> pure <$> compileInstance ToDefinition def
-        (DefaultPragma _     , _      , Axiom{}   ) -> compilePostulate def
-        (DefaultPragma _     , _      , Function{}) -> compileFun True def
-        (DefaultPragma ds    , _      , Record{}  ) -> pure <$> compileRecord (ToRecord False ds) def
+      isInstance <- case defInstance def of
+        Just cl -> isClassName cl
+        Nothing -> return False
```
flupe, line 85: Could you use `anyM` from `Agda.Utils.Monad`?

```suggestion
      isInstance <- anyM isClassName (defInstance def)
```

src/Agda2Hs/Compile/Term.hs
```diff
@@ -272,7 +272,7 @@ specialClassFunction3 v f = specialClassFunction $ \case
   es               -> v `eApp` es
 
 fromNat :: ProjCompileRule
-fromNat = specialClassFunction1 (hsVar "fromIntegral") $ \case
+fromNat = specialClassFunction2 (hsVar "fromIntegral") $ \v _ -> case v of
```
flupe, line 275: I trust that the code still works, but can you explain what requires using `specialClassFunction2` now?

I couldn't find where the signature of `fromIntegeral` got changed, thus where the extra argument comes from.

src/Agda2Hs/Compile/Utils.hs
```diff
@@ -165,15 +165,24 @@ isClassModule m
...
-      Right Defn{defName = r, theDef = Record{}} ->
-        -- It would be nicer if we remembered this from when we looked at the record the first time.
-        processPragma r <&> \case
-          ClassPragma _       -> True
-          ExistingClassPragma -> True
-          _                   -> False
-      _                       -> return False
+    if Just m == minRec then return True else isClassName (mnameToQName m)
+
+isClassName :: QName -> C Bool
```
flupe, line 170: The name kinda gives it away, but can you please add a Hoogle comment?

src/Agda2Hs/Compile/Utils.hs
```diff
@@ -165,15 +165,24 @@ isClassModule m
...
+isClassName q = getConstInfo' q >>= \case
+  Right Defn{defName = r, theDef = Record{}} ->
+    -- It would be nicer if we remembered this from when we looked at the record the first time.
+    processPragma r <&> \case
+      ClassPragma _       -> True
+      ExistingClassPragma -> True
+      _                   -> False
+  _                       -> return False
+
+isClassType :: Type -> C Bool
```
flupe, line 180: Same here.

flupe (approved):
Very nice!

I think it closes #218, perhaps even #251.

src/Agda2Hs/Compile/Term.hs
jespercockx: The extra argument is the (instance) argument `{{_ : Constraint n}}` in the type of `fromNat`:

https://github.com/agda/agda/blob/0def25833eb92660499799280aa791f58253b4bb/src/data/lib/prim/Agda/Builtin/FromNat.agda#L11

Previously this was treated as a class constraint (i.e. checked for canonicity and then dropped) but now it becomes are regular argument instead so we need to drop it by hand. In an ideal world, we would update the type of `fromNat` to mark the constraint as erased, but it is defined in `Agda.Builtin` so changing it is not easy.

flupe: Do you think defining our own `fromNat` and `Number` class is a bad idea?

src/Agda2Hs/Compile/Utils.hs
jespercockx: We don't really have Haddock comments for *any* of the functions, so just adding the one here feels like it would be a bit out of place.

src/Agda2Hs/Compile/Term.hs
jespercockx: If you want to use literals on the Agda side then it *has* to use the `fromNat`/`fromNeg`/`fromString` functions from the builtin Agda modules.

jespercockx: Well yes you can declare the `BUILTIN` yourself, but declaring builtins is something we've been avoiding so far to preserve compatibility with other libraries (at least for the erased parts of your `agda2hs` code).

PR #292, opened 2024-02-28
Very small enhancements to existing modules in Haskell.Prim

viktorcsimma:
Actually, I've added two missing fixities and a `fromMaybe` function; and a comment to Haskell.Prim.String. So nothing special.

### DIFF
--- a/lib/Haskell/Prim/String.agda
+++ b/lib/Haskell/Prim/String.agda
@@ -7,10 +7,10 @@ open import Haskell.Prim.Foldable
 
 --------------------------------------------------
 -- String
--- This is _not_ the builtin String type of Agda;
+-- This is _not_ the builtin String type of Agda
 -- which is defined by postulates.
--- fromString can be used to produce Haskell strings
--- from that.
+-- `fromString` can be used to convert back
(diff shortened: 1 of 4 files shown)


lib/Haskell/Prim/String.agda
```diff
@@ -7,6 +7,10 @@ open import Haskell.Prim.Foldable
 
 --------------------------------------------------
 -- String
+-- This is _not_ the builtin String type of Agda;
+-- which is defined by postulates.
+-- fromString can be used to produce Haskell strings
+-- from that.
```
omelkonian, line 13: Nice addition, but needs rephrasing:
1. No need for `;`, just say "This is _not_ the builtin String type of Agda that is defined by postulates".
2. `fromString` actually gives *Agda* strings, not Haskell ones, and in fact this type of strings in this module is what corresponds to Haskell strings. Thus, rephrase as "`fromString` can be used to convert back to builtin Agda strings"?

omelkonian (changes requested):
Some minor improvements I would like before merging.

viktorcsimma:
Done; thank you:)

PR #297, opened 2024-03-01
Adding mechanism to compile Haskell.Data.* modules to Data.* modules

viktorcsimma:
I've opened a separate PR for this. There is now a list of module prefixes which get compiled to corresponding Haskell modules. E.g. `Haskell.Data.Map` becomes `Data.Map`.

Actually, the test files used depend on maps; so I can only add them after they get merged (I'll include them in the Map PR)

### DIFF
--- a/src/Agda2Hs/Compile/Name.hs
+++ b/src/Agda2Hs/Compile/Name.hs
@@ -166,7 +166,7 @@ compileQName f
 
     -- Prefix of modules that correspond to similarly named Haskell modules
     -- (e.g. "Haskell.Data.Map" is "Data.Map").
-    hsModules = ["Haskell.Data"]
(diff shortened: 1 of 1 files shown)


viktorcsimma:
Now added Haskell.Control, too (that is going to be important for #298).

omelkonian:
Have you considered extending the "rewrite-rules" mechanism for providing these translation? IMO this is preferable to polluting the code base with such ad-hoc configurations.

viktorcsimma:
I think it would be better for it to work out-of-the-box as the user expects them. But in fact, we could include some kind of "default" configuration file to which we move both this and `defaultSpecialRules` in Name.hs. Do you think about something like this, or do I miss the point?

PR #304, opened 2024-03-08
Type-based compilation of everything + fix for #264

### LINKED ISSUE #264: Internal error with erased module parameter and 'where' function
The following file produces an internal error when given to `agda2hs`:
```agda
module ReduceBug {@0 name  : Set} where

postulate 
  Term : Set → Set

reduce : Term name → Term name
reduce v = go v
  where postulate
    go : Term name → Term name

{-# COMPILE AGDA2HS reduce #-}
```
The error message:
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Substitute.hs:230:38 in Agda-2.6.4.1-8d2624b65fdd0db0ceced7a4414cd73aa757a2fddd00ce30a24ac87f9baa99a8:Agda.TypeChecking.Substitute
```
and the last debug output before it dies:
```
Compiling function:  ReduceBug.go
Compiling top-level type {@0 name = name₁ : Set} (v : Term name₁) →
                         Term name₁ → Term name₁
Getting telescope for ReduceBug._
Current context:  [(v, Term v), {(name₁, Set)}, {(name, Set)}]
Context arguments:  [{_}, {_}, v]
Module section:  {@0 name = name₂ : Set} (v₁ : Term name₂)
```
@flupe You added the support for erased module parameters. I looked at the code for `compileTopLevelType` but could not really make sense of it. Could you please take a look?

jespercockx:
This PR continues on the path that @flupe has started with https://github.com/agda/agda2hs/pull/270 and pushes type-based compilation to every corner of `agda2hs`. It also fixes #264 by removing the horrible `compileTopLevelType` function and introducing a better way to handle explicit `forall`s.

Ignoring a number of refactorings, this PR makes the following semantic changes to `agda2hs`:
- Parameters of data and record types and of type aliases are now compiled according to their type rather than their hiding/quantity annotations. In particular, hidden data parameters are now supported.
- Instance declarations are now also compiled according to their type, so non-erased hidden fields are now supported.
- The check that unboxed records have a single non-erased field is now type-based, so in particular (visible) fields of type `Level`, `Size`, and any type in `Prop` are now allowed (and dropped during compilation).
- Non-erased module parameters are now allowed. Type parameters are turned into explicit `forall`s to each top-level function in the module, while other parameters are compiled as normal arguments.

The test suite passes and I have added some minimal extra tests, but it could definitely still use some more testing, so please let me know if you have any examples you'd like me to try.

### DIFF
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -157,9 +157,6 @@ compileFun' withSig def@Defn{..} = do
     m = qnameModule defName
     n = qnameName defName
     x = hsName $ prettyShow n
-    endsInSort t = do
-      TelV tel b <- telView t
-      addContext tel $ ifIsSort b (\_ -> return True) (return False)
     err = "Not supported: type definition with `where` clauses"
 
 
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -133,7 +133,6 @@ compileFun' withSig def@Defn{..} = do
 
       compileTopLevelType withSig defType $ \ty -> do
 
-        let filtered = filter keepClause funClauses
         weAreOnTop <- isJust <$> liftTCM  (currentModule >>= isTopLevelModule)
         pars <- getContextArgs
 
@@ -141,7 +140,7 @@ compileFun' withSig def@Defn{..} = do
         -- if the current module isn't the toplevel module
         unless weAreOnTop $
           reportSDoc "agda2hs.compile.type" 8 $ "Applying module parameters to clauses: " <+> prettyTCM pars
-        let clauses = if weAreOnTop then filtered else filtered `apply` pars
+        let clauses = if weAreOnTop then funClauses else funClauses `apply` pars
 
         typ <- if weAreOnTop then pure defType else piApplyM defType pars
 
@@ -166,16 +165,15 @@ compileClause curModule mproj x t c =
     compileClause' curModule mproj x t c
 
 compileClause' :: ModuleName -> Maybe QName -> Hs.Name () -> Type -> Clause -> C (Maybe (Hs.Match ()))
-compileClause' curModule projName x _  c@Clause{clauseBody = Nothing} = pure Nothing
 compileClause' curModule projName x ty c@Clause{..} = do
   reportSDoc "agda2hs.compile" 7  $ "compiling clause: " <+> prettyTCM c
-  reportSDoc "agda2hs.compile" 17 $ "Old context: "      <+> (prettyTCM =<< getContext)
-  reportSDoc "agda2hs.compile" 17 $ "Clause telescope: " <+> prettyTCM clauseTel
-  reportSDoc "agda2hs.compile" 17 $ "Clause type:      " <+> prettyTCM clauseType
-  reportSDoc "agda2hs.compile" 17 $ "Function type:    " <+> prettyTCM ty
-  reportSDoc "agda2hs.compile" 17 $ "Clause patterns:  " <+> text (prettyShow namedClausePats)
 
-  addContext (KeepNames clauseTel) $ do
+  ifNotM (keepClause c) (pure Nothing) $ addContext (KeepNames clauseTel) $ do
+    reportSDoc "agda2hs.compile" 17 $ "Old context: "      <+> (inTopContext . prettyTCM =<< getContext)
+    reportSDoc "agda2hs.compile" 17 $ "Clause telescope: " <+> inTopContext (prettyTCM clauseTel)
+    reportSDoc "agda2hs.compile" 17 $ "Clause type:      " <+> prettyTCM clauseType
+    reportSDoc "agda2hs.compile" 17 $ "Function type:    " <+> prettyTCM ty
+    reportSDoc "agda2hs.compile" 17 $ "Clause patterns:  " <+> text (prettyShow namedClausePats)
 
     toDrop <- case projName of
       Nothing -> pure 0
@@ -212,11 +210,22 @@ compileClause' curModule projName x ty c@Clause{..} = do
 
     let rhs = Hs.UnGuardedRhs () hsBody
         whereBinds | null whereDecls = Nothing
-                   | otherwise       = Just $ Hs.BDecls () (concat whereDecls)
+                  | otherwise       = Just $ Hs.BDecls () (concat whereDecls)
         match = case (x, ps) of
           (Hs.Symbol{}, p : q : ps) -> Hs.InfixMatch () p x (q : ps) rhs whereBinds
           _                         -> Hs.Match () x ps rhs whereBinds
     return $ Just match
+
+  where
+    keepClause :: Clause -> C Bool
+    keepClause c@Clause{..} = case (clauseBody, clauseType) of
+      (Nothing, _) -> pure False
+      (_, Nothing) -> pure False
+      (Just body, Just cty) -> compileDom (domFromArg cty) >>= \case
+        DODropped  -> pure False
+        DOInstance -> pure False -- not __IMPOSSIBLE__ because of current hacky implementation of `compileInstanceClause`
+        DOType     -> __IMPOSSIBLE__
+        DOTerm     -> pure True
 
 
 -- TODO(flupe): projection-like definitions are missing the first (variable) patterns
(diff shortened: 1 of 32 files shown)


jespercockx:
One current limitation is that you cannot have `where` declarations with `forall`s in their type signature. In fact, if you try to do that by using a module inside your `where` block, the definitions in that block simply get dropped. For example,
```agda
module _ {a : Set} where
  f1 : a → a
  f1 x = f2 id
    where module _ {b : Set} where
      f2 : (a → b) → b
      f2 g = f3 (g x)
        where module _ {c : Set} where
          f3 : c → c
          f3 x = x
      
{-# COMPILE AGDA2HS f1 #-}
```
is compiled to the rather sad
```haskell
f1 : forall a . a -> a
f1 x = f2 id
```
The reason why the definitions of `f2` and `f3` are dropped is nothing new: it is a consequence of our imperfect attempt at reconstructing the information of which definitions belong to which clause. We use the function `isFatherModuleOf` (https://github.com/agda/agda2hs/blob/0611268835b7314fd58a6fad30a2b05418422e5f/src/Agda2Hs/AgdaUtils.hs#L41-L42) which only considers *direct* submodules but not indirect ones.

src/Agda2Hs/Compile/Function.hs
```diff
@@ -215,12 +213,22 @@ compileClause' curModule projName x ty c@Clause{..} = do
 
     let rhs = Hs.UnGuardedRhs () hsBody
         whereBinds | null whereDecls = Nothing
-                   | otherwise       = Just $ Hs.BDecls () (concat whereDecls)
+                  | otherwise       = Just $ Hs.BDecls () (concat whereDecls)
```
flupe, line 216: can you bring back the space? 👀

src/Agda2Hs/Compile/Function.hs
```diff
@@ -215,12 +213,22 @@ compileClause' curModule projName x ty c@Clause{..} = do
...
 
+keepClause :: Clause -> C Bool
+keepClause c@Clause{..} = case (clauseBody, clauseType) of
+  (Nothing, _) -> pure False
+  (_, Nothing) -> pure False
+  (Just body, Just cty) -> compileDom (domFromArg cty) >>= \case
+    DODropped  -> pure False
+    DOInstance -> pure False -- not __IMPOSSIBLE__ because of current hacky implementation of `compileInstanceClause`
+    DOType     -> __IMPOSSIBLE__
+    DOTerm     -> pure True
```
flupe, line 230: Can replace `>>=` with `<&>` and drop the `pure`s, probably.

src/Agda2Hs/Compile/Record.hs
```diff
@@ -168,16 +170,30 @@ compileRecord target def = do
...
+      =<< text "Unboxed record" <+> prettyTCM (defName def)
+      <+> text "should have exactly one non-erased field"
+
+  where
+    nonErasedFields :: Telescope -> C [String]
+    nonErasedFields EmptyTel = return []
+    nonErasedFields (ExtendTel a tel) = compileDom a >>= \case
+      DODropped  -> underAbstraction a tel nonErasedFields
+      DOType -> genericDocError =<< text "Type field in unboxed record not supported"
+      DOInstance -> genericDocError =<< text "Instance field in unboxed record not supported"
```
flupe, line 198: Do we not? I assumed since #284 instance search can be used to generate Haskell terms, wouldn't this also work fine for instance fields in unboxed records? Don't know if it'd ever useful to define an unboxed record so that the field is always resolved by instance search, but anyways.

Unless instance arguments for non-class records already produce a `DOTerm` output? Would make sense.

src/Agda2Hs/Compile/Type.hs
```diff
@@ -155,42 +107,67 @@ compileType t = do
...
+        <+> text "in Haskell type"
+      vs <- compileTypeArgs (snd $ unDom xi) args
       x  <- hsName <$> compileDBVar x
       return $ tApp (Hs.TyVar () x) vs
 
     Sort s -> return (Hs.TyStar ())
 
-    Lam argInfo restAbs
-      | not (keepArg argInfo) -> underAbstraction_ restAbs compileType
+    -- This case is only reachable for erased lambdas
```
flupe, line 128: Is this true?
I need to test locally but say, if the type is `(f : (Type -> Type) -> Type) (x : f (\y -> Nat)) -> Nat` or smthg, wouldn't the lambda be allowed without the check?

flupe (approved):
Very nice! I have more questions than remarks really, have yet to think of more tests to add.

src/Agda2Hs/Compile/Type.hs
```diff
@@ -155,42 +107,67 @@ compileType t = do
...
          | otherwise -> fail
 
     Var x es | Just args <- allApplyElims es -> do
-      vs <- compileTypeArgs args
+      xi <- lookupBV x
+      unless (usableModality xi) $ genericDocError
+        =<< text "Cannot use erased variable" <+> prettyTCM (var x)
+        <+> text "in Haskell type"
+      vs <- compileTypeArgs (snd $ unDom xi) args
       x  <- hsName <$> compileDBVar x
```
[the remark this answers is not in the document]
jespercockx: `compileDBVar` is used in other places where it's not necessary to check the modality, so I decided to move it here. Since it is just changing where the check happens, I don't think this should cause any breaking change, but if you have a counterexample I'm happy to investigate.

src/Agda2Hs/Compile/Function.hs
```diff
@@ -116,50 +115,50 @@ compileFun withSig def@Defn{..} =
     $ compileFun' withSig def
 
 -- inherit existing (instantiated) locals
-compileFun' withSig def@Defn{..} = do
+compileFun' withSig def@Defn{..} = inTopContext $ do
```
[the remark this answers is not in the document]
jespercockx: It's mostly because of calls to `compileFun'` for compiling pattern-matching lambdas, which can appear in arbitrary contexts but should really be compiled in the top context. For `where` clauses we also call it from the context of the clause, which should (ideally) be the same as the context of the definition's module but we have to be careful not to add it twice. But I'll see if I can add the call to `inTopContext` to those places so we can avoid the blanket call here.

src/Agda2Hs/Compile/Function.hs
```diff
@@ -116,50 +115,50 @@ compileFun withSig def@Defn{..} =
...
+      typ <- piApplyM defType pars
+      reportSDoc "agda2hs.compile.type" 8 $ "Function type (after instantiation): " <+> prettyTCM typ 
 
-        -- We only instantiate the clauses to the current module parameters
-        -- if the current module isn't the toplevel module
-        unless weAreOnTop $
-          reportSDoc "agda2hs.compile.type" 8 $ "Applying module parameters to clauses: " <+> prettyTCM pars
-        let clauses = if weAreOnTop then filtered else filtered `apply` pars
+      sig <- if not withSig then return [] else do
+        checkValidFunName x
```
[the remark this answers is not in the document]
jespercockx: This might be a nice refactoring to do in future work ;)

src/Agda2Hs/Compile/Record.hs
jespercockx: I did not break any test cases so the feature was never really supported :)

jespercockx: > Unless instance arguments for non-class records already produce a DOTerm output? Would make sense.

This is indeed the case.

src/Agda2Hs/Compile/Type.hs
jespercockx: You are right, this is actually a real problem.
```agda
foo : (f : (Set → Set) → Set) (x : f (λ y → Nat)) (y : f List) → Nat
foo f x y = 42

{-# COMPILE AGDA2HS foo #-}
```
is now compiled to 
```haskell
foo :: f Natural -> f [] -> Natural
foo x y = 42
```
where `f` is applied to two arguments of a different kind. I'll add the check again that we don't compile non-erased lambdas.

src/Agda2Hs/Compile/Function.hs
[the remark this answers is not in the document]
jespercockx: Issue: https://github.com/agda/agda2hs/issues/306

Issue #153, opened 2023-02-28
Improve agda2hs' test infrastructure


jespercockx:
Currently the test infrastructure for agda2hs is stupid: for each test we need to add an import to `AllTests.hs` or else the test is not actually run. Instead we could adapt Agda's testing infrastructure using `tasty-silver` to make running the tests less error-prone.

andreasabel:
Task description:

Make a golden value testsuite for `agda2hs` by converting the `Makefile`-based testsuite in `test/` into a Haskell one based on the `tasty-golden` package.  The testsuite should be runnable via `cabal test`.

The final directory structure for the test files should be:

- successful tests in `test/Succeed`: move there the `.agda` files from `test/` and `test/Cubical/`

- failing tests in `test/Fail`

Each successful test is based on an `.agda` file which:

1. is compiled using `agda2hs` to a `.hs` file which is the golden value
2. the generated `.hs` is compiled with `ghc` and this should succeed.

Each failing test is a `.agda` file which should fail agda2hs: the reported error is the golden value stored in a `.err` file.  The golden value may not contain absolute paths so they should be relative or relativized to make the testsuite portable.

Golden values are stored beside along their `.agda` file.

The compilation of the `.agda` tests need `test/agda2hs-test.agda-lib` and `lib/base` structure it references.

The test runner should build the TestTree by looking for `.agda` files in `test/Succeed` or `test/Fail` resp.

The tests and their golden values should also be `data-files` in the `.cabal` file (using wildcards).

PR #307, opened 2024-03-15
Fix for issue #305 + some related problems

### LINKED ISSUE #305: Typeclass functions in parametrized modules do not work
Even with my PR https://github.com/agda/agda2hs/pull/304, there still seems to be a problem with the compilation of typeclass instances in parametrized modules (whether the parameters are erased or not). Here's an example:

```agda
open import Haskell.Prelude

module InstanceBug (@0 X : Set) where

record Class (a : Set) : Set where
  field
    foo : a → a
open Class {{...}} public

{-# COMPILE AGDA2HS Class class #-}

instance
  ClassInt : Class Int
  ClassInt .foo = _+ 1

{-# COMPILE AGDA2HS ClassInt #-}

test : Int
test = foo 41

{-# COMPILE AGDA2HS test #-}
```
The error message:
```
No instance of type Class Int was found in scope.
```
With `-vtc.instance:50` we get an idea of the reason:
```
checkCandidateForMeta
  t    = Class Int
  t'   = (@0 X : Set) → Class Int
  term = ClassInt
instance search: checking (@0 X : Set) → Class Int <= Class Int
} (exception)
candidate failed type check: /home/jesper/agda-libs/agda-core/src/Agda/Core/InstanceBug.agda:19,1-5
                             (@0 X : Set) → Class Int !=< Class Int
```
The module parameter `X` has become an explicit argument to the instance, which means it does not play well with the check we do for canonicity.

jespercockx:
This fixes #305 by setting the right checkpoint for Agda so that when we call instance search for checking canonicity, the instances are correctly applied to the module parameters.

I also found and fixed a related regression introduced in my PR https://github.com/agda/agda2hs/pull/304 that caused us to no longer check the canonicity of superclass instances when defining a subclass instance.

When I run the test suite locally on this branch, there is something weird that happens with the golden files for two modules:

- `NonCanonicalSuperclass.agda` produces an unexpected success despite producing the correct error when running `agda2hs` on it directly
- `StreamFusion.agda` I did not touch in any way, but suddenly appears in a different location

I feel like our test suite is getting increasingly flaky and hard to manage, so perhaps we should look into improving it soon (https://github.com/agda/agda2hs/issues/153)

### DIFF
--- a/src/Agda2Hs/Compile.hs
+++ b/src/Agda2Hs/Compile.hs
@@ -82,6 +82,8 @@ compile opts tlm _ def =
 
       isInstance <- anyM (defInstance def) isClassName 
 
+      reportSDoc "agda2hs.compile" 15  $ text "Is instance?" <+> prettyTCM isInstance
+
       case (p , theDef def) of
         (NoPragma            , _         ) -> return []
         (ExistingClassPragma , _         ) -> return []
--- a/src/Agda2Hs/Compile/ClassInstance.hs
+++ b/src/Agda2Hs/Compile/ClassInstance.hs
@@ -217,7 +217,9 @@ compileInstanceClause' curModule ty (p:ps) c
           body <- case clauseBody c' of
             Nothing -> genericDocError =<< text "not allowed: absurd clause for superclass"
             Just b  -> return b
-          checkInstance body
+          addContext (clauseTel c) $ do
+            liftTCM $ setModuleCheckpoint curModule
+            checkInstance body
           reportSDoc "agda2hs.compile.instance" 20 $ vcat
             [ text "compileInstanceClause dropping clause"
             , nest 2 $ prettyTCM c
--- a/src/Agda2Hs/Compile/ClassInstance.hs
+++ b/src/Agda2Hs/Compile/ClassInstance.hs
@@ -84,7 +84,7 @@ compileInstance ToDefinition def@Defn{..} =
       let mod = qnameModule defName
       reportSDoc "agda2hs.compile.instance" 25 $ text "compileInstance module: " <+> prettyTCM mod
       (ds, rs) <- concatUnzip
-              <$> mapM (compileInstanceClause (qnameModule defName) defType) funClauses
+              <$> mapM (\c -> withClauseLocals mod c $ compileInstanceClause mod defType c) funClauses
       reportSDoc "agda2hs.compile.instance" 20 $ vcat $
         text "compileInstance compiled clauses: " :
         map (nest 2 . text . pp) ds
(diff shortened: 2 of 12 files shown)


omelkonian:
> StreamFusion.agda I did not touch in any way, but suddenly appears in a different location

c.f. the changes of 416234a81e36c, where I moved the cubical tests under `Cubical` for a more consistent test folder structure

jespercockx:
> c.f. the changes of https://github.com/agda/agda2hs/commit/416234a81e36c6604564b964da127210bb036188, where I moved the cubical tests under Cubical for a more consistent test folder structure

I get that you changed it, but I don't understand why `StreamFusion.hs` still appears in the `golden` directory rather than in `golden/Cubical`.

jespercockx:
I think that was it: after manually deleting `test/agda2hs` the discrepancies suddenly went away. Though now there are *still* CI errors that I cannot produce locally. The mystery hunt continues!

jespercockx:
After exorcising the ghosts from my repository I managed to fix the remaining internal errors over the weekend. However, the issue *still* remains when a class function is used inside a pattern-matching lambda:
```agda
module _ (@0 X : Set) where

record Class (a : Set) : Set where
  field
    foo : a → a
open Class {{...}} public

{-# COMPILE AGDA2HS Class class #-}

instance
  ClassInt : Class Int
  ClassInt .foo = _+ 1

{-# COMPILE AGDA2HS ClassInt #-}

yetAnotherTest : Int
yetAnotherTest = case Just True of λ where
  Nothing → error "unreachable"
  (Just y) → foo 5
{-# COMPILE AGDA2HS yetAnotherTest #-}
```
still gives the error:
```
No instance of type Class Int was found in scope.
```

jespercockx:
I got one step further again: the checkpoint is now being set in the context of the module rather than the context of an individual clause, which means the example above now works. However, the problems are still not over as the check still fails when the instance is defined in a *different* module that is applied and opened in the current module:
```agda
module M1 (@0 X : Set) where

  record Class (a : Set) : Set where
    field
      foo : a → a
  open Class {{...}} public

  {-# COMPILE AGDA2HS Class class #-}

  instance
    ClassInt : Class Int
    ClassInt .foo = _+ 1

  {-# COMPILE AGDA2HS ClassInt #-}

module M2 (@0 X : Set) where

  open M1 X

  tester : Int
  tester = foo 41

  {-# COMPILE AGDA2HS tester #-}
```
Error is the same as before:
```
No instance of type M1.Class X Int was found in scope.
```
I guess the problem is that we are using the instance table of the top-level module rather than the one that was used inside the module `M2`, and this top-level instance table does not include the instances from locally opened modules.

omelkonian:
Let's see if we can use `Origin::UserWritten` to only run the consistency check in these cases, thus allowing us not to re-typecheck and re-construct everything that Agda has done before.

jespercockx:
> Let's see if we can use Origin::UserWritten to only run the consistency check in these cases, thus allowing us not to re-typecheck and re-construct everything that Agda has done before.

Unfortunately this does not work: in Agda's internal syntax, calls to type class functions are represented as a `Proj` elimination, which does not include any `ArgInfo` with the `Origin`. It does have a `ProjOrigin`, but this seems to be set to `ProjPrefix` whether or not the instance was written by hand so it is useless for our purpose here. So unfortunately it seems we do not have the information available to determine whether the argument was provided manually or not.

(As an aside: Agda's pretty-printer seems to struggle with a similar problem, as it prints `foo 41` here as `Class.foo ClassInt (int64 41)`.)

jespercockx:
Since this PR constitutes a clear improvement to the status quo and I don't have a solution for the remaining problem, I believe it's best to merge this now and leave adding support for using instances coming from module applications as a separate issue.

PR #313, opened 2024-04-12
[ fix #302 ] Move _/=_ function out of the Eq record

### LINKED ISSUE #302: _/=_ generates incorrect `iEq…`
```agda
not0 : Int → Bool
not0 n = n /= 0
{-# COMPILE AGDA2HS not0 #-}
```
generates
```haskell
import qualified Prelude ((/=))

not0 :: Int -> Bool
not0 n = (Prelude./=) iEqInt n 0
```
where `iEqInt` is an instance from our `Haskell.Prim.Eq`, not a Haskell function, and, importing `Prelude` `qualified`ly breaks all mentions to it (such as `Int` and `Bool` from that generated snippet)

jespercockx:


omelkonian:
@jespercockx The golden file is missing; feel free to cherry-pick the fix from the `issue302-fix` branch: https://github.com/agda/agda2hs/commit/a9d79b81299e4ea245634b561f1b7e320d67aa83

flupe:
If we want to stay faithful to [the Haskell interface](https://hackage.haskell.org/package/base-4.20.0.0/docs/Data-Eq.html), shouldn't `_/=
_` be turned into a record field instead?

jespercockx:
> If we want to stay faithful to [the Haskell interface](https://hackage.haskell.org/package/base-4.20.0.0/docs/Data-Eq.html), shouldn't _/= _ be turned into a record field instead?

Ideally yes, but (1) it would break current code and (2) there seem to be some plans to move `(/=)` out of the type class in the Haskell prelude.

PR #325, opened 2024-05-13
Makefile robustness

jakobn-ai:
### Update make test trigger

`test/agda2hs` was rebuilt upon changes in `src/*.hs` only, but these are not all the source files (in fact, they're the less frequently used ones). Extend this to all files.
We can't use a Makefile wildcard here because the nest goes deeper.

### Update test make dependencies

for efficient and correct multiprocessing


jespercockx:
Thank you! This was indeed bothering me for a while already.

PR #326, opened 2024-05-23
Rebuild test files when appropriate

jakobn-ai:
of course it wasn't all -.- sorry for the noise

- Rebuild positive also upon new `agda2hs` binary
- Rebuild negative on new `agda2hs` binary, removing `force-recompile`


jespercockx:
Thanks!

Issue #125, opened 2022-11-14
Check that constructors are not overloaded


jespercockx:
Haskell does not have overloaded constructors, so we should make sure that we do not allow several constructors with the same name to be compiled to Haskell.

jespercockx:
Fixed in #328

PR #328, opened 2024-06-08
Add error for overloaded Haskell constructors

PPKFS:
Solves #125. 

Unfortunately, it seems the Agda compiler only does its compiler magic for each definition independently, so there is no state carried over as to the previous definitions that have been compiled.

We check all the constructors generated for duplicates prior to `writeOutput` in `postCompile` and refuse to continue if there are duplicates. This does also add scaffolding for future post-compilation, pre-render verification steps.


jespercockx:
> Unfortunately, it seems the Agda compiler only does its compiler magic for each definition independently, so there is no state carried over as to the previous definitions that have been compiled.

Yes, we have discussed this limitation before, and then we came to the conclusion that the best way to have some shared state would be to use an `IORef` (which is possible since the `C` monad implements `MonadIO`).

However, for this issue the solution you propose also seems like a reasonable approach.

src/Agda2Hs/Compile.hs
```diff
@@ -106,3 +107,25 @@ compile opts tlm _ def =
...
+            Hs.DataDecl _ _ _ _ cons _ <- a
+            Hs.QualConDecl _ _ _ con <- cons
+            return $ case con of
+              Hs.ConDecl _ n _ -> n
+              Hs.InfixConDecl _ _ n _ -> n
+              Hs.RecDecl _ n _ -> n
+          duplicateCons = filter ((> 1) . length) . group . sort  $ allCons
+      when (length duplicateCons > 0) $
+        genericDocError =<< vcat (map (\x -> text $ "Cannot generate multiple constructors with the same identifier: " <> Hs.prettyPrint (head x)) duplicateCons)
+      return (length duplicateCons == 0)
```
jespercockx, line 131: Please add a newline at EOF!

test/Fail/Issue125.agda
```diff
@@ -0,0 +1,16 @@
...
+
+data B : Set where
+  ACtr : B
+
+{-# COMPILE AGDA2HS B #-}
+
+data C : Set where
+  Ca : C
+
+{-# COMPILE AGDA2HS C #-}
```
jespercockx, line 16: Same here

src/Main.hs
```diff
@@ -71,7 +71,7 @@ backend = Backend'
   , preCompile            = checkConfig
   , postCompile           = \ _ _ _ -> return ()
   , preModule             = moduleSetup
-  , postModule            = writeModule
+  , postModule            = \opts env isM m out -> verifyOutput opts env isM m out >> writeModule opts env isM m out
```
jespercockx, line 74: Maybe you could just make `verifyOutput` a part of `writeModule`, or else add a new function that does these two functions in sequence?

jespercockx:
The test suite fails on the following:
```
== Comparing output ==
diff -r build/Issue125.err golden/Issue125.err
1c1
< Cannot generate multiple constructors with the same identifier: ACtr
---
> Found multiple declarations of constructor: ACtr
```
Please run `make test` and `make golden` and then commit the changes to the golden files.

src/Main.hs
PPKFS: I think a new function is probably best. Having `writeModule`, in `Render`, also do final verification feels like a weird mixup of concerns

PPKFS:
I was looking for a `cabal test` equivalent and didn't think to check for a makefile. Oops.

Issue #108, opened 2022-09-17
Add lawful subclasses of existing type classes


jespercockx:
It would be good to have standard definitions of the type class laws for the classes in the Prelude, and prove them for the instances we provide. Here is an example of how this would look:
```agda
open import Haskell.Prelude

record LawfulMonad (m : Set → Set) {{iMonad : Monad m}} : Set₁ where
  field
    left-id  : ∀ {a b} (x : a) (f : a → m b)
             → (return x >>= f) ≡ f x
    right-id : ∀ {a} (k : m a)
             → (k >>= return) ≡ k
    assoc    : ∀ {a b c} (k : m a) (f : a → m b) (g : b → m c)
             → ((k >>= f) >>= g) ≡ (k >>= (λ x → f x >>= g))
open LawfulMonad

instance
  _ : LawfulMonad Maybe
  _ = λ where
    .left-id → λ x f → refl
    .right-id Nothing → refl
    .right-id (Just x) → refl
    .assoc Nothing f g → refl
    .assoc (Just x) f g → refl
```

Classes for which we should define laws:
- [ ] `Eq` and `Ord`
- [ ] `Num`
- [ ] `BoundedBelow`, `BoundedAbove`, and `Bounded`
- [ ] `Enum`
- [ ] `Semigroup` and `Monoid`
- [ ] `Functor`, `Applicative`, and `Monad`
- [ ] `Foldable` and `Traversable`

pmbittner:
As part of PR #331, I implemented the Laws and instances for the `Num` typeclass. Sorry for spamming the issue feed on accident here with all my rebases. I will take better care with my commit messages / rebases in the future.

jespercockx:
Since a lot of type classes have been added now, it seems it would be easier to use separate issues to track the remaining ones that are still missing. So I'll go ahead and close this one. Please create new issues if you find any laws that are missing!

PR #331, opened 2024-06-08
Implementing `IsLawfulNum` as part of Issue #108

pmbittner:
This pull-request adds the laws for the `Num` typeclass in Haskell as part of issue #108.

Progress:
- [x] formalized the laws for Num typeclass
- [x] `Nat` instance
- [x] `Int` instance
- [x] `Integer` instance
- [x] `Word` instance
- ~`Double` instance~

There was one law we cannot formalize yet, which is `fromInteger (toInteger i) == i` because there is no `toInteger` function available yet. I documented this law as missing.

For the `Nat` instance, I mostly copied code from the Agda standard library for comparability, consistency, and also because it is easy :sweat_smile:.

For `Integer` I had to come up with new proofs because the definitions for this type are slightly different from those in the standard library. I oriented on the proofs from the standard library where possible though. Also I tried to favor equational reasoning in favor of `rewrite` for better readability.

Arithmetics of `Word` is mostly relying on arithmetics of `Nat` and arithmetics of `Int` is defined solely in terms of arithmetics on `Word`. I hence opted to reuse the laws from the other instances, respectively. In fact, reducing the arithmetics for a `Num` type in terms of another `Num` instance is feasible as long as this reduction is homomorphic.

To formalize this observation, I added definitions for homomorphisms, embeddings, as well as some other common predicates in a new module `Haskell.Prim.Function` with respective proofs in `Haskell.Law.Function`. I thought that such definitions might be useful in other contexts as well, if desired. Using these definitions, we can then define the circumstances under which we may conclude an `ILawfulNum` instance from another one.

The question remains whether such abstractions are desired. On the one hand, these are quite common concepts. On the other hand, they introduce another layer of indirection and abstractions, making some of the proofs harder to comprehend. Also, should we reuse the definitions such as `Associative`, `Commutative`, and so on in the definition of `IsLawfulNum` or should we inline those (as it currently is) for readability?

Moreover, I had to introduce six new axioms to prove lawfulness of `Word`, all of which are stated in `lib/Haskell/Law/Num/Word.agda`. I documented the new axioms and explained why they should be reasonable. It would be good if these could be reviewed. I am not 100% sure that they are correct and minimal.

Also, the `Double` instance cannot be proven as of now. It is solely based on primitive functions for which no axioms exist yet. So I left that out for now.


lib/Haskell/Law/Num/Nat.agda
```diff
@@ -0,0 +1,37 @@
+module Haskell.Law.Num.Nat where
+
+open import Haskell.Prim
+open import Haskell.Prim.Num
+
+open import Haskell.Law.Equality
+open import Haskell.Law.Num.Def
+
+-- Where to put this?
```
jespercockx, line 9: I think it's fine to put it here until we find a better place for it.

lib/Haskell/Law/Num/Nat.agda
```diff
@@ -0,0 +1,37 @@
...
+  Nat-+-idʳ : ∀ (x : Nat) → x + 0 ≡ x
+  Nat-+-idʳ zero    = refl
+  Nat-+-idʳ (suc x) = cong suc (Nat-+-idʳ x)
+
+open Termination
+
+instance
+  iLawfulNum : IsLawfulNum Nat
+  IsLawfulNum.+-assoc iLawfulNum = {!!}
+  IsLawfulNum.+-comm iLawfulNum zero y = sym (Nat-+-idʳ y)
```
jespercockx, line 25: Style-wise, I would avoid inlining the proofs into the instance definition. We could instead give each property a top-level name, or we could define the whole record as a record type and put each of the fields as a definition in a `where` block.

jespercockx (commented):
This looks like a great start! I've left two small comments.

lib/Haskell/Law/Num/Nat.agda
pmbittner: Ok, then I will also gather all the other definitions here for now.

lib/Haskell/Law/Num/Nat.agda
pmbittner: Sounds good to me. This will also resolve the termination problems I faced so far. :)

pmbittner:
I noticed that I had to make `negNat` and `subNat` public. `Integer` arithmetics heavily rely on these functions and I could not prove properties about said arithmetics as long as these functions are opaque (i.e., cannot be inlined).

We could stay with `negNat` and `subNat` being private but we would have to move the various lemmas to the same file. I guess that would be worse though than just making the functions public (for now).

pmbittner:
I started working on the instance for `Word`. Given that words are represented by primitive conversions to  and from `Nat`, I aimed to reuse the `Nat` instance for `IsLawfulNum`. My idea was to determine the missing axioms for Word's primitive functions, and then conclude the `IsLawfulNum` instance in a generic way. Only after having formalized what it takes to derive an `IsLawfulNum` instance from another, I noticed that this would not be possible for `Word` from `Nat`.

The problem is that Words are modelled as a true subset of the Nats but
Nats do not support negation nor minus but Words do (with modulo).
This makes it impossible to reuse most proofs of `iLawfulNumNat`.
I do not know whether it would make sense to represent Words as a subset of Integers instead?

lib/Haskell/Law/Num/Word.agda
```diff
@@ -0,0 +1,98 @@
...
+mulWord-distributeʳ-addWord = map-distributeʳ addWord addNat mulWord mulNat w2n n2w n2w∘w2n≡id w2n-+-hom w2n-*-hom mulNat-distributeʳ-addNat
+
+-- TODO: I have trouble figuring out negation of words and how
+--       to prove the following. Can we prove it?
+neg-inv-word : ∀ (x : Word) → addWord x (negateWord x) ≡ 0
+neg-inv-word x =
+  x + negateWord x                                                ≡⟨⟩
+  n2w (w2n x + w2n (negateWord x))                                ≡⟨⟩
+  n2w (w2n x + w2n (n2w (monusNat 18446744073709551616 (w2n x)))) ≡⟨ {!!} ⟩
+  0 ∎
```
pmbittner, line 83: There is a single hole left. I was not able to find a proof here yet. We might have to introduce additional axioms on `Word` but I am not sure.

pmbittner: fixed

pmbittner:
This PR is now ready to review.

jespercockx (approved):
Great work, thank you very much!

PR #332, opened 2024-06-09
[ fix #119 ] Raise a warning when an unknown name appears in generated Haskell code

### LINKED ISSUE #119: Check that all used definitions are actually defined in the Haskell code
@flupe mentioned to me that it would be a good idea to check that all defined symbols that are used in the generated Haskell code are also actually defined in the Haskell code. This would require some kind of "foreign" pragma for annotating a function that is already defined in the Haskell code.

jespercockx:
This fixes #119 by raising a warning when an unknown name appears in the generated Haskell code. The following names are considered as "known":
- Names that have a COMPILE pragma
- Names that have an associated rewrite rule in the agda2hs config
- Names that are defined in a primitive module (i.e. anything in the `agda2hs` library)
- Names of record fields that represent a type class
- Names of constructors of datatypes that have a COMPILE pragma

For the test suite, I didn't figure out yet how to capture the warning message. At the moment, the test just produces "unexpected success". If anyone has an idea how to set up the test for the warning, that's very welcome. (Otherwise I'll take a look at how Agda itself does it).


jespercockx:
Since 1.3 has been released, I will go ahead and merge this now so it can get some testing before the next release.

Issue #105, opened 2022-09-17
Add support for user-definable tuples (WAS: User-definable functions for builtin Haskell syntax)


jespercockx:
It would be nice if we could make some of the functions that are currently defined in the Prelude (such as `if_then_else_`) user-definable instead, and perhaps even allow multiple Agda functions that compile to the same Haskell syntax. Here is a list of some functions for which this could be done:

- [x] Tuple constructors `_,_`, `_,_,_`, ...
- [ ] `if_then_else_`
- [ ] `case_of_`
- [ ] Integer and string literals
- [ ] `enumFromThenTo`
- [ ] ...

jespercockx:
See also #74 for some motivation

jespercockx:
The types of `if_then_else_` and `case_of_` are now more general so there's less need for having multiple versions of them, and for literals and ranges the need to have multiple versions of them is a bit dubious too. So since we have the most important part done, I will close this now so it can be added to the 1.3 milestone.

PR #342, opened 2024-07-23
Add support for custom tuples

jespercockx:
This PR addresses the first point in #105 by adding support for compiling an Agda record type to a Haskell tuple. The feature is designed to be pretty flexible, so you can compile any record type and `agda2hs` will just take the (compiled) field types as the tuple components. It also allows defining *unboxed* tuples, which was previously not supported by `agda2hs`. However, there are no checks in place that the [rules for unboxed tuples](https://downloads.haskell.org/~ghc/6.12.1/docs/html/users_guide/primitives.html#unboxed-tuples) are followed, so it might generate invalid Haskell code if used improperly.

### DIFF
--- a/agda2hs.cabal
+++ b/agda2hs.cabal
@@ -86,4 +86,5 @@ executable agda2hs
                        ViewPatterns
                        NamedFieldPuns
                        PatternSynonyms
+                       NondecreasingIndentation
   ghc-options:         -Werror -rtsopts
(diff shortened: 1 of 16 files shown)


jespercockx:
At the moment, tuple projections are not yet supported, and I'm not sure what's the best way to handle them. Here are two options:

1. Generate projection functions for each projection of the record type (but then there will be many redundant copies of the projections if you have a lot of tuple types)
2. Compile projections to built-in functions `fst` and `snd` (but what to do about tuples with more than 2 components?)

A hybrid (2 for binary tuples, 1 for everything else) is also possible in theory.

jespercockx:
One more option I just thought of: take option 1 except that for binary tuple types, we check that the projections are called `fst` and `snd`, and we don't generate any code in that case.

PR #358, opened 2024-09-07
Fix for #357

### LINKED ISSUE #357: Levels should not always be erased
The interaction of levels and polymorphism pose problems, as demonstrated by the following example:

```agda
open import Haskell.Prelude
open import Agda.Primitive using (lzero)

k : a → b → a
k x _ = x
{-# COMPILE AGDA2HS k #-}

testK : Nat
testK = k 42 lzero
{-# COMPILE AGDA2HS testK #-}
```

In the use site we erase the level argument completely, resulting in the wrong number of arguments when applying the polymorphic `k` function:
```hs

k :: a -> b -> a
k x _ = x

testK :: Natural
testK = k 42
```
```
    • Couldn't match expected type ‘Natural’
                  with actual type ‘b0 -> a0’
    • Probable cause: ‘k’ is applied to too few arguments
      In the expression: k 42
      In an equation for ‘testK’: testK = k 42
  |
9 | testK = k 42
  |         ^^^^

```

jespercockx:
This fixes #357 by adding a check that type arguments are valid Haskell types before dropping them. It also makes some changes to the Prelude (in particular to `guard`) so it conforms to this new check.

### DIFF
--- a/lib/Haskell/Extra/Erase.agda
+++ b/lib/Haskell/Extra/Erase.agda
@@ -14,10 +14,12 @@ module Haskell.Extra.Erase where
     @0 xs  : List a
 
   record Erase (@0 a : Set ℓ) : Set ℓ where
-    constructor Erased
-    field @0 get : a
+    instance constructor iErased
+    field @0 {{get}} : a
   open Erase public
   {-# COMPILE AGDA2HS Erase tuple #-}
+
+  pattern Erased x = iErased {{x}}
 
   infixr 4 ⟨_⟩_
   record Σ0 (@0 a : Set) (b : @0 a → Set) : Set where
--- a/lib/Haskell/Control/Monad.agda
+++ b/lib/Haskell/Control/Monad.agda
@@ -4,7 +4,8 @@ open import Haskell.Prim
 open import Haskell.Prim.Bool
 open import Haskell.Prim.Monad
 open import Haskell.Prim.String
+open import Haskell.Extra.Erase
 
-guard : {{ MonadFail m }} → (b : Bool) → m (b ≡ True)
-guard True = return refl
+guard : {{ MonadFail m }} → (b : Bool) → m (Erase (b ≡ True))
+guard True = return (Erased refl)
 guard False = fail "Guard was not True"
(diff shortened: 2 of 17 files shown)


lib/Haskell/Extra/Erase.agda
```diff
@@ -14,8 +14,8 @@ module Haskell.Extra.Erase where
     @0 xs  : List a
 
   record Erase (@0 a : Set ℓ) : Set ℓ where
-    constructor Erased
-    field @0 get : a
+    instance constructor Erased
+    field @0 {{get}} : a
```
flupe, line 18: What's the motivation for changing the interface of `Erase` and making the field an instance? Seems like it forces explicit brackets `{{}}` everywhere.

flupe (approved):
Looks good! The errors are perhaps not ideal to get what's wrong but I understand it would be annoying to make them clearer.

lib/Haskell/Extra/Erase.agda
jespercockx: Well, it was to make sure that instance search could actually find an element of type `Erase A` when there is an erased instance of type `A` in scope. But I found a better way to get the same effect, which does not break backwards compatibility (see newly pushed version).

flupe: Is it a situation that arises often? Having instances of `a` laying around? For equality proofs I get, but otherwise not so sure. But I trust you that the interface hasn't changed :). (I would have kept the old `Erase` definition and defined `iErased` separately but it probably doesn't matter much. And with yours one can decide to match and directly introduce an instance, which I don't know if you can do otherwise)

jespercockx: The opposite way (defining `Erased` as the constructor and `iErased` as the pattern synonym) is not accepted by Agda:
```
.../agda2hs/lib/Haskell/Extra/Erase.agda:22,21-22: error: [IllegalInstanceVariableInPatternSynonym]
Variable is bound as instance in pattern synonym, but does not
resolve as instance in pattern: x
when scope checking the declaration
  pattern iErased ⦃ x ⦄ = Erased x
  ```

PR #363, opened 2024-09-18
Properly differentiate between multiple anonymous modules 

### LINKED ISSUE #353: Calling between functions in (multiple) anonymous modules causes arguments to disappear
This Agda code
```agda
open import Haskell.Prelude

module MultipleAnonModules where

module _ (a : Bool) (b : Int) where
    foo : Bool
    foo = not a
{-# COMPILE AGDA2HS neg #-}

module _ (b : Int) where
    bar : Bool
    bar = foo True 2 -- Note that foo is called with two arguments

{-# COMPILE AGDA2HS bar #-}
```
generates this Haskell code
```haskell
module MultipleAnonModules where

foo :: Bool -> Int -> Bool
foo a b = not a

bar :: Int -> Bool
bar b = foo 2 -- Here the first argument disappeared
```
For each parameter in the second anonymous module, one argument in the call into the first module disappears. It doesn't matter what the parameters are or if they are used or not.

-----
The same issue appears if both functions are in the same module:
```agda
open import Haskell.Prelude
module OneAnonModule where

module _ (a : Bool) (b : Int) where
    foo : Bool
    foo = not a

    bar : Bool
    bar = foo

    {-# COMPILE AGDA2HS bar #-}
    {-# COMPILE AGDA2HS neg #-}
```
This generates
```haskell
module OneAnonModule where

foo :: Bool -> Int -> Bool
foo a b = not a

bar :: Bool -> Int -> Bool
bar a b = foo
```
But this instance of the bug feels slightly less surprising, since now the generated code looks like the original.

anka-213:
<s>Partial fix for issue</s> Fixes #353.

Instead of checking if the prettyprinted qualified module name is a prefix to determine if a module is contained within another, we check if the actual list of `Name` is a prefix, which uses the internal hashes.

With this fix, calling another function within the same local module will still cause an issue, but at least calling other modules won't cause arguments to magically disappear.

I am also not completely sure if the direction of the check is correct? Wouldn't we want to skip arguments only if calling a function from a surrounding module, rather than only when calling into a submodule? The current version is mostly equivalent to an equality check, since you can't call into a where-clause from outside of it

### DIFF
--- a/src/Agda2Hs/Compile.hs
+++ b/src/Agda2Hs/Compile.hs
@@ -37,6 +37,7 @@ initCompileEnv tlm rewrites = CompileEnv
   , minRecordName     = Nothing
   , locals            = []
   , compilingLocal    = False
+  , whereModules      = []
   , copatternsEnabled = False
   , rewrites          = rewrites
   }
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -228,14 +228,14 @@ compileClause' curModule projName x ty c@Clause{..} = do
     -- `where` declarations (if there are any) in order to drop the arguments
     -- that correspond to the pattern variables of this clause from the calls to
     -- the functions defined in the `where` block.
-    let inWhereModule = case children of
-          []    -> id
-          (c:_) -> withCurrentModule $ qnameModule c
+    -- let inWhereModule = case children of
+    --       []    -> id
+    --       (c:_) -> withCurrentModule $ qnameModule c
 
     let Just body            = clauseBody
         Just (unArg -> typ)  = clauseType
 
-    hsBody <- inWhereModule $ compileTerm typ body
+    hsBody <- compileTerm typ body
 
     let rhs = Hs.UnGuardedRhs () hsBody
         whereBinds | null whereDecls = Nothing
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -222,20 +222,16 @@ compileClause' curModule projName x ty c@Clause{..} = do
           /\ (curModule `isFatherModuleOf`) . qnameModule
 
     children   <- filter isWhereDecl <$> asks locals
-    whereDecls <- compileLocal $ mapM (getConstInfo >=> compileFun' True) children
-
-    -- Jesper, 2023-10-30: We should compile the body in the module of the
-    -- `where` declarations (if there are any) in order to drop the arguments
-    -- that correspond to the pattern variables of this clause from the calls to
-    -- the functions defined in the `where` block.
-    -- let inWhereModule = case children of
-    --       []    -> id
-    --       (c:_) -> withCurrentModule $ qnameModule c
+    -- TODO: remove this when Agda exposes where-provenance in 'Internal' syntax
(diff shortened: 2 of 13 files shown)


anka-213:
I tested calling a function from an outer where-clause from a nested where-clause and it did indeed have the expected issue. However, the fix is not as simple as flipping the order of the arguments, since we get the number of arguments from the current module instead of the module where the function is defined. I don't yet know how to get the info from another module that you just have the name of.

```agda
nested : Bool → Nat
nested b = inner1
  where
    outer : Nat
    outer = 5

    inner1 : Nat
    inner1 = 2 + inner2
      where
        inner2 : Nat
        inner2 = outer
```
generates
```haskell
nested :: Bool -> Natural
nested b = inner1
  where
    outer :: Natural
    outer = 5
    inner1 :: Natural
    inner1 = 2 + inner2
      where
        inner2 :: Natural
        inner2 = outer b -- Notice the extra b argument here
```
with the current version.

jespercockx:
Thank you for the PR, this is great! For the remaining issue, I wonder if we could throw an error when compiling a call to a function from the current module (or a nested one), that module is anonymous and parametrized, and it is not the `where` module we are currently compiling. But perhaps getting access to all the information for those checks is tricky.

anka-213:
my guess is that it is no more difficult to solve the remaining issue than it is to show that error

anka-213:
this version of the code covers almost all cases, including calling between functions in the same module.
```haskell
    defIsClass <- isClassModule defMod -- If we're compiling a minimal class
    -- TODO: simplify this when Agda exposes where-provenance in 'Internal' syntax
    isLocal <- asks compilingLocal -- or if we're calling from a where-clause
    localDefs <- asks locals -- or if we're calling into a where-clause

    (ty', args') <-
      -- and we're calling into a module that we're currently in, then we drop the module parameters.
      if f `elem` localDefs (isLocal || defIsClass) && (mnameToList defMod `isPrefixOf` mnameToList currentMod) then do
```

 the only issue that remains is if you call a function in the same local module, from within a where-clause, we currently have no easy way to distinguish it from a binding in an outer where-clause (until Agda exposes where-provenance in 'Internal' syntax), so that call still loses arguments:

```agda
module _ (b : Bool) where

    bar : Bool
    bar = True
    {-# COMPILE AGDA2HS bar #-}

module _ (b : Bool) where

    foo : Bool
    foo = True
    {-# COMPILE AGDA2HS foo #-}

    -- Works fine now
    bar : Bool
    bar = foo
    {-# COMPILE AGDA2HS bar #-}

    -- The call to foo won't get the necessary Bool argument here
    callFromNested : Bool
    callFromNested = nested
      where nested = foo
    {-# COMPILE AGDA2HS callFromNested #-}
```

anka-213:
Now I've added code that keeps track of which where-clauses are currently in scope, so we can use that to determine if the arguments should be removed or not. This change replaced the compilation of the body in the module of the where-declaration, since we keep track of it in a different way now.
https://github.com/agda/agda2hs/blob/078622c928dbf36c0ecbb70143b37780a40dc09a/src/Agda2Hs/Compile/Function.hs#L227-L230

This should fix the last issues. All test cases succeed now!

jespercockx:
Awesome! Thanks again.

PR #365, opened 2024-09-22
Don't crash on parse errors in deriving clauses

### LINKED ISSUE #351: Parse errors on deriving clauses a class itself causes a crash
I was trying to add deriving to all my data types with a naive regex which caused the following code:
```agda
record Example : Set where
{-# COMPILE AGDA2HS Example deriving (Show, Eq, Ord) class #-}
```
Trying to run agda2hs on it gives the following error:
```
$ agda2hs Bug.agda
Checking Bug (Bug.agda).
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/Utils/FileName.hs:56:20 in Agda-2.6.4.1-7xG50m7UQLK4krJEY6TPg:Agda.Utils.FileName
```

Not that anyone should do this, but the error message said it's a bug, so I might as well report it so mistakes like these can get better error messages.

anka-213:
It turns out that if you don't specify a file-name, then the Haskell parser [haskell-src-exts uses the filename `<unknown>.hs`](https://hackage.haskell.org/package/haskell-src-exts-1.23.1/docs/src/Language.Haskell.Exts.ParseMonad.html#defaultParseMode),
which we afterwards try to parse as an absolute filepath to extract the location of the error, which causes a crash.

Since the location from the Haskell parser wouldn't be accurate anyways, we change it to just report the entire pragma as the location of the error.

This PR also makes the code for converting Haskell locations into Agda locations return `Nothing` for the filename, instead of crashing if there is no valid filename in the error. This might not be desired since we should be able to provide a more helpful filename and location and this could make it easier to miss the issues with bad error reporting. That change currently has no impact, since this PR also prevents it from triggering.

Fixes #351

### DIFF
--- a/src/Agda2Hs/HsUtils.hs
+++ b/src/Agda2Hs/HsUtils.hs
@@ -17,6 +17,8 @@ import Agda.Syntax.Position
 import Agda.Utils.FileName ( mkAbsolute )
 import Agda.Utils.List ( initLast )
 import Agda.Utils.Maybe.Strict ( toStrict )
+import System.FilePath (isAbsolute)
+import Agda.Utils.Maybe (boolToMaybe)
 
 -- Names ------------------------------------------------------------------
 
@@ -199,7 +201,7 @@ cloc (ConName l _) = l
 
 srcSpanToRange :: SrcSpan -> Range
 srcSpanToRange (SrcSpan file l1 c1 l2 c2) =
-  intervalToRange (toStrict $ Just $ mkRangeFile (mkAbsolute file) Nothing) $ Interval (pos l1 c1) (pos l2 c2)
+  intervalToRange (toStrict $ boolToMaybe (isAbsolute file) $ mkRangeFile (mkAbsolute file) Nothing) $ Interval (pos l1 c1) (pos l2 c2)
   where pos l c = Pn () 0 (fromIntegral l) (fromIntegral c)
 
 srcLocToRange :: SrcLoc -> Range
--- a/src/Agda2Hs/Pragma.hs
+++ b/src/Agda2Hs/Pragma.hs
@@ -71,13 +71,13 @@ parseStrategy _          = Nothing
 newtypePragma :: String
 newtypePragma = "newtype"
 
-processDeriving :: String -> ([Hs.Deriving ()] -> ParsedPragma) -> C ParsedPragma
-processDeriving d pragma =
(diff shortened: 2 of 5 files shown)


anka-213:
It would also be possible to adjust the haskell parser to have a more accurate location, using similar code to the one used for foreign pragmas:
https://github.com/agda/agda2hs/blob/22f37f57fe5c17193147d90f548ecf327b5a694c/src/Agda2Hs/Pragma.hs#L38-L45

The issue with that is that we don't know exactly how far into the line the `deriving` keyword is, so it's difficult to adjust the column number accurately in the haskell parser.

It might however be worthwhile to provide the list of extensions to the haskell parser for deriving clauses, like the `FOREIGN` parser does, so it can accurately handle things like "deriving via", but that's a separate unrelated issue that would require more refactoring.

jespercockx:
Thanks for the PR! I generally agree that having a more precise error location would be nice but here the issue seems rather minor so I would propose just merging this PR as is.

jespercockx (approved):

anka-213:
I agree. I believe that getting the more precise location reliably is currently blocked by getting more range information from the Agda parser. 

More specifically: We only get the range of the full pragma, not the range of the string contents we get from it, so the best we could do is to guess that there is only one whitespace between each keyword and calculate how long the `{-# COMPILE AGDA2HS FunctionName` prefix _should be_ and then subtract that or make the dummy data type declaration have the exact same length, so that the haskell error message gets an accurate location _most of the time_.

jespercockx:
Ok, then let's go ahead and merge this.

anka-213:
@jespercockx btw, on a related note: the code currently doesn't report an error on invalid keywords after `{-# COMPILE AGDA2HS FunctionName`, but instead just ignores them. Is this intended behavior?

In the function
https://github.com/agda/agda2hs/blob/22f37f57fe5c17193147d90f548ecf327b5a694c/src/Agda2Hs/Pragma.hs#L85-L102
The final line ignores errors:
https://github.com/agda/agda2hs/blob/22f37f57fe5c17193147d90f548ecf327b5a694c/src/Agda2Hs/Pragma.hs#L102

This means that this is accepted:
```agda
record Foo : Set where
{-# COMPILE AGDA2HS Foo "&"€%(€&( #-}
```

However, getting precise error location for this would have the same issue, so we would have to either give an imprecise error or make a guess (that will be true 99% of the time). But I guess the lack of precision doesn't matter too much there either.

jespercockx:
I'm not sure if it's intentional or not, I think it's been like that since the original implementation by @UlfNorell

PR #368, opened 2024-09-24
Add support for GHC 9.10.2

jespercockx:
This adds support for GHC 9.10.2, and drops support for older versions of GHC (< 9.4.8) from CI to avoid having too many different CI runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]   # macOS-latest, windows-latest
-        cabal: [3.8]
-        ghc: [8.8.4, 8.10.7, 9.2.5, 9.4.3, 9.6.3]
+        cabal: [3.10.3]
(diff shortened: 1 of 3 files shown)


jespercockx:
One point of discussion: if people agree with dropping CI for GHC versions prior to 9.4.8, should we also increase the `base` dependency from 4.13 to 4.17 (the version used by GHC 9.4.*)? In other words, is it better to be loose in the version bounds so people can easily build with older GHCs, or strict so that we can guarantee that building will actually work?

jespercockx:
I merged this for now, we can always bump the version bound later.

UlfNorell:
Usually we base our decisions of which versions to support on what gets shipped with Ubuntu LTS or some such. What's the status of ghc-9.4 there?

jespercockx:
9.4.8 is the version of ghc currently recommended by ghcup. Since ghcup works so well nowadays, I felt like that was more relevant than whatever is in the Ubuntu LTS.

For reference, Ubuntu 20.04 has ghc 8.8.1, 22.04 has 8.8.4, and 24.04 has 9.4.7

PR #372, opened 2024-10-01
[ fix #370 ] add locate command, data-files in cabal config

### LINKED ISSUE #370: Add agda2hs library to data-files + add command to print location of the library
As suggested by @HeinrichApfelmus in https://github.com/agda/agda2hs/issues/366#issuecomment-2377083047 and https://github.com/agda/agda2hs/issues/366#issuecomment-2377411491

flupe:
Added the `data-files` info to bundle the agda2hs prelude with the cabal package.
Had to bump cabal version to `2.4` for the wildcards to be accepted.

`agda2hs locate` will now print out the path to the bundled `.agda-lib` file of the prelude, still need to update the documentation to explain how to run the command when setting up agda2hs.

Let me know if this command should be renamed to something else (I just chose the same name as `agda-mode locate`).

### DIFF
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -21,7 +21,7 @@ import Agda2Hs.Config ( checkConfig )
 import Agda2Hs.Compile.Types
 import Agda2Hs.Render
 
-import Paths_agda2hs ( version )
+import Paths_agda2hs ( version, getDataFileName )
 
 
 -- | Agda2Hs default config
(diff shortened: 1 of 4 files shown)


src/Main.hs
```diff
@@ -87,8 +87,15 @@ isInteractive = do
...
-  -- Issue #201: disable backend when run in interactive mode
-  isInt <- isInteractive
-  if isInt
-    then runAgda [Backend backend{isEnabled = const False}]
-    else runAgda [Backend backend]
+  args <- getArgs
+
+  -- Issue #370: `agda2hs locate` will print out the path to the prelude agda-lib file
+  if args == ["locate"] then 
+    getDataFileName "agda2hs.agda-lib" >>= putStrLn
```
omelkonian, line 94: I suggest using `putStrLn =<< getDataFileName "agda2hs.agda-lib"` for readability.

omelkonian (approved):
LGTM; I agree with naming the command `locate` to be consistent with normal Agda.

jespercockx (approved):

jespercockx:
Thank you, this looks great!

anka-213:
hmm, I wonder if this change would affect the nix-distribution, as it packages the agda2hs library separately. I'll have to research that

Issue #366, opened 2024-09-23
Release agda2hs 1.3


jespercockx:
It seems like we're long overdue for a new release (as indicated by the comment by @HeinrichApfelmus in https://github.com/agda/agda2hs/issues/361#issuecomment-2364047031_).

One thing that might be worth doing is to investigate some lower/upper bounds on the dependencies (see e.g. https://github.com/agda/agda2hs/issues/347 and https://github.com/agda/agda2hs/pull/350#issuecomment-2363512179).

jespercockx:
I've uploaded a release candidate to https://hackage.haskell.org/package/agda2hs-1.3/candidate

It would be great if people could give it a test run (perhaps @omelkonian @HeinrichApfelmus @anka-213). If there are no complaints I'll do a proper release next week.

HeinrichApfelmus:
> It would be great if people could give it a test run

Do you have a git commit that this release corresponds to?

I need to test the combination of `agda2hs` executable and `agda2hs.agda-lib` Agda library. If I understand correctly, only the executable is release on Hackage? 🤔 I'm asking for a git commit, because that would allow me to get the combination as presented in a specific `flake.nix`.

jespercockx:
It should correspond to https://github.com/agda/agda2hs/commit/770f209c3e1aa94ba9d34c1488379e83190d589d

jespercockx:
> If I understand correctly, only the executable is released on Hackage?

I think that's the case, indeed. I'm not sure if there's an easy way to release the library also on Hackage, do you think it's worth investigating or is it fine to just distribute it via github?

HeinrichApfelmus:
>  do you think it's worth investigating or is it fine to just distribute it via github?

Distribution via Github, specifically the `flake.nix` works for me.

For a "proper" release that is accessible to a larger audience than just me, I think that investing time into creating release artifacts is unavoidable. I suppose that

* adding the library to the `data-files:` in `agda2hs.cabal`
* having the executable `agda2hs` use the bundled library by default, with a command-line option to use the traditional (if clumsy) `AGDA_LIB` way of specifying package dependencies.

would do the trick.

More generally, in order to make `agda2hs` "the missing proof assistant for Haskell", there is probably no way around integrating with `.cabal`.

jespercockx:
[...]

PR #374, opened 2024-10-01
[ fix #373 ] enable existing-class on postulated type formers

### LINKED ISSUE #373: Allow `existing-class` for postulated type classes
We should enable the `existing-class` pragma for postulated type classes:

```agda
postulate
  MyShow : Set -> Set
  myshow : {{ MyShow a }} -> a -> String
{-# COMPILE MyShow existing-class #-}

ok : {{ MyShow a }} -> a -> String
ok = myshow
```

Currently `MyShow` is not considered a class (since #284), and thus the instance arguments get turned into regular Haskell arguments and not class constraints:

```haskell
ok : MyShow a -> a -> String
ok x = myshow x
```

flupe:
This should fix [the problem highlighted here](https://github.com/agda/agda2hs/issues/366#issuecomment-2385463366).

We may want to add an additional check so that only postulates with the right type signatures can use `existing-class` (i.e proper type former, no class constraints).

### DIFF
--- a/src/Agda2Hs/Compile/Utils.hs
+++ b/src/Agda2Hs/Compile/Utils.hs
@@ -173,6 +173,7 @@ isClassName q = getConstInfo' q >>= \case
       ClassPragma _       -> True
       ExistingClassPragma -> True
       _                   -> False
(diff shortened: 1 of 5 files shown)


flupe:
Closing this, since I can't think of any situations where it isn't preferable to introduce a record and use existing-class.

Issue #316, opened 2024-05-08
Better support for wrapping Haskell modules via `postulate`


HeinrichApfelmus:
## Problem statement

The mission statement of Adga2hs is to carve out a sublanguage of Agda that translates directly to Haskell, so that we can prove properties about Haskell-in-Agda programs in Agda. This is highly useful for producing high assurance Haskell code. Personally, I think of Agda2hs as "the missing proof assistant for Haskell".

I particularly enjoy that Agda2hs allows me to develop proofs about Haskell programs in an **incremental fashion**: I can write and prove selected parts of my program in Haskell-in-Agda, **export** them to Haskell, and use the result together with existing Haskell code that was only tested, not proven.

However, I feel that this incremental approach would become even more powerful if Agda2hs better supported the **import** of Haskell code. Here, I don't mean a syntactic import of Haskell code (that would be amenable to proof), but an import of Haskell functions and "tested" properties via Agda's `postulate` mechanism. Let me call this "wrapping a Haskell module".

Put differently, at the moment, Agda2hs implements a translation from `.agda` to `.hs` files where the `.agda` file is the "source of truth". I would like to see a translation from `.agda` to `.hs` where the `.hs` file acts as a "source of truth".

## Proposed solution: `COMPILE … IMPORT` pragma

Agda2hs has made a design decision to translate `.agda` files to `.hs` files. In particular, it does not attempt to parse Haskell files. This design works particularly well for using Agda2hs as a preprocessor for Haskell. In order to explore this point in the design space properly, we need to take it to its logical extreme before changing it. Hence, I'll leave the Haskell to the Haskell compiler.

Consider the following module, which is legal Adga2hs today:

```agda
module Haskell.Cardano.Crypto.Wallet where

{-# FOREIGN AGDA2HS
{-# LANGUAGE UnicodeSyntax #-}
import qualified Cardano.Crypto.Wallet as CC
#-}

postulate
  XPub : Set  -- plaintext public key
  XPrv : Set  -- plaintext private key

  toXPub : XPrv → XPub

{-# FOREIGN AGDA2HS
type XPub = CC.XPub
type XPrv = CC.XPrv

toXPub :: XPrv → XPub
toXPub = CC.toXPub
#-}
```

[...]

PR #379, opened 2024-11-15
Always drop "Haskell." prefix from module names

jespercockx:
This supersedes https://github.com/agda/agda2hs/pull/297. The idea (due to @omelkonian) is that any module starting with `Haskell.` should be treated as defining part of the FFI with Haskell and should thus not be checked for having COMPILE pragmas, but instead be assumed to exist on the Haskell side while dropping the `Haskell.` prefix from the module name.


jespercockx:
Two things to consider before merging this:
- There is a risk of introducing a name collision between `Haskell.Foo.Bar` and `Foo.Bar`. Can we do anything about this?
- Should we enforce that `Haskell.` modules do not produce any output?
- Can we make it clearer to the user that defining a `Haskell.` module is an unsafe thing to do? We can certainly add a note to the documentation, but is there anything else we can do?

jespercockx:
Related: https://github.com/agda/agda2hs/issues/316

viktorcsimma:
This is a nice idea:) I often utilise similar concepts, and this would be of great help.

* For the first one: I have checked this and GHC, in this case, prefers the user-defined module. So `Foo.Bar` would trump `Haskell.Foo.Bar` in this case. But I do not have an idea on how to circumvent this, apart from a note in the docs.
* For the second: I think we should. If the output were `Foo.Bar`, it would trump the Haskell builtin version, which we probably do not want. If it were `Haskell.Foo.Bar`, it would be an ordinary agda2hs-generated module.
* For the third: what about hiding the feature behind a flag? Like with `--erasure` in Agda.

jespercockx:
Regarding the first two: actually, if we implement the second and enforce no Haskell code being generated from `Haskell.` files, then there is no risk of conflicts. There is one module in our own prelude that does not conform to this restriction (https://github.com/agda/agda2hs/blob/master/lib/Haskell/Extra/Loop.agda), but I don't think anyone is using it so we could just drop it.

I like the idea of having a pragma that should be in every module under `Haskell.`, that way we can prevent users from accidentally running into this feature.

jespercockx:
I added the check that modules under the `Haskell.` namespace do not produce any Haskell output, solving the first two problems. I don't think that 3 is such a big problem anymore, since the Haskell code will not compile anyway unless the corresponding Haskell bindings are already there. So let's just leave it as it is now, and add the flag if it proves to be really necessary.

PR #382, opened 2024-12-18
Make the nix development compatible with flake-less nix

liesnikov:
This only changes the internal arrangement of files, allowing folks who don't use flakes to build agda2hs.
The interface exposed to those who use flakes should stay the same.


liesnikov:
Running some test builds [here](https://github.com/liesnikov/scope/actions/runs/12397718771) and [here](https://github.com/liesnikov/agda-core/actions/runs/12397832144), both seem okay. (the second one errors out the same way as before and after all library managment has been done)

PR #396, opened 2025-01-24
Add `Haskell.Data.List`

HeinrichApfelmus:
The purpose of this pull request is to add some functions from the module `Data.List`. Specifically, we add

* `nub` with a proof of a property `prop-elem-nub`
* `deleteAll` as a helper function for `nub`
* `sort` as a postulate

We also add `Haskell.Data.Ord`, this is needed for `sortOn`.

I quite like the property

```agda
prop-elem-nub
  : ∀ ⦃ _ : Eq a ⦄ ⦃ _ : IsLawfulEq a ⦄
      (x : a) (ys : List a)
  → elem x (nub ys)
    ≡ elem x ys
```

as it expresses a logical equivalence in terms of `≡`, I think that this is only available for `Bool`, not for `Prop` / `Set`. The proof is also very succinct.

On the other hand, the proof for `prop-elem-deleteAll` is somewhat lengthy, but I'm not sure that this can be avoided, as we have to use transitivity of `==` at crucial points in the proof. I am using the induction hypothesis `prop-elem-deleteAll x y zs` in a `with` clause in order to avoid doing substitutions by hand via `cong`.

### DIFF
--- a/lib/Haskell/Data/List.agda
+++ b/lib/Haskell/Data/List.agda
@@ -4,12 +4,27 @@ open import Haskell.Prelude
 
 open import Haskell.Data.Ord using (comparing)
 
+open import Haskell.Law.Eq
+open import Haskell.Law.Equality
+
 {-----------------------------------------------------------------------------
     Operations
 ------------------------------------------------------------------------------}
 
 partition : (a → Bool) → List a → (List a × List a)
 partition p xs = (filter p xs , filter (not ∘ p) xs)
+
+-- | Delete all occurrences of an item.
+-- Not part of 'Data.List'.
+deleteAll : ⦃ _ : Eq a ⦄ → @0 ⦃ IsLawfulEq a ⦄ → a → List a → List a
+deleteAll x = filter (not ∘ (x ==_))
+
+-- | These semantics of 'nub' assume that the 'Eq' instance
+-- is lawful.
+-- These semantics are inefficient, but good for proofs.
+nub : ⦃ _ : Eq a ⦄ → @0 ⦃ IsLawfulEq a ⦄ → List a → List a
+nub [] = []
+nub (x ∷ xs) = x ∷ deleteAll x (nub xs)
 
 postulate
   sortBy : (a → a → Ordering) → List a → List a
--- a/lib/Haskell/Data/List.agda
+++ b/lib/Haskell/Data/List.agda
@@ -2,6 +2,8 @@ module Haskell.Data.List where
 
 open import Haskell.Prelude
 
+open import Haskell.Data.Ord using (comparing)
+
 {-----------------------------------------------------------------------------
     Operations
 ------------------------------------------------------------------------------}
@@ -9,3 +11,14 @@ open import Haskell.Prelude
 partition : (a → Bool) → List a → (List a × List a)
 partition p xs = (filter p xs , filter (not ∘ p) xs)
 
+postulate
+  sortBy : (a → a → Ordering) → List a → List a
+
+sort : ⦃ Ord a ⦄ → List a → List a
+sort = sortBy compare
+
+sortOn : ⦃ Ord b ⦄ → (a → b) → List a → List a
+sortOn f =
+    map snd
+    ∘ sortBy (comparing fst)
+    ∘ map (λ x → let y = f x in seq y (y , x))
--- a/lib/Haskell/Data/List.agda
+++ b/lib/Haskell/Data/List.agda
@@ -37,3 +37,54 @@ sortOn f =
     map snd
     ∘ sortBy (comparing fst)
     ∘ map (λ x → let y = f x in seq y (y , x))
+
+{-----------------------------------------------------------------------------
+    Properties
+------------------------------------------------------------------------------}
+
+-- | A deleted item is no longer an element.
+--
+prop-elem-deleteAll
+  : ∀ ⦃ _ : Eq a ⦄ ⦃ _ : IsLawfulEq a ⦄
+      (x y : a) (zs : List a)
+  → elem x (deleteAll y zs)
+    ≡ (if x == y then False else elem x zs)
+--
+prop-elem-deleteAll x y []
+  with x == y
+... | False = refl
+... | True  = refl
+prop-elem-deleteAll x y (z ∷ zs)
+  with recurse ← prop-elem-deleteAll x y zs
+  with y == z in eqyz
+... | True
+    with x == z in eqxz
+...   | True
+      rewrite equality' _ _ (trans (equality x z eqxz) (sym (equality y z eqyz)))
+      = recurse
+...   | False
+      = recurse
+prop-elem-deleteAll x y (z ∷ zs)
+    | False
+    with x == z in eqxz
+...   | True
+      rewrite equality x z eqxz | eqSymmetry y z | eqyz
+      = refl
+...   | False
+      = recurse
+
+-- | An item is an element of the 'nub' iff it is
+-- an element of the original list.
+--
+prop-elem-nub
+  : ∀ ⦃ _ : Eq a ⦄ ⦃ _ : IsLawfulEq a ⦄
+      (x : a) (ys : List a)
+  → elem x (nub ys)
+    ≡ elem x ys
+--
+prop-elem-nub x [] = refl
+prop-elem-nub x (y ∷ ys)
+  rewrite prop-elem-deleteAll x y (nub ys)
+  with x == y
+... | True = refl
+... | False = prop-elem-nub x ys
--- a/lib/Haskell/Data/List.agda
+++ b/lib/Haskell/Data/List.agda
@@ -1,0 +1,11 @@
+module Haskell.Data.List where
+
+open import Haskell.Prelude
+
+{-----------------------------------------------------------------------------
+    Operations
+------------------------------------------------------------------------------}
+
+partition : (a → Bool) → List a → (List a × List a)
+partition p xs = (filter p xs , filter (not ∘ p) xs)
+
--- a/lib/Haskell/Data/Ord.agda
+++ b/lib/Haskell/Data/Ord.agda
@@ -1,0 +1,9 @@
+module Haskell.Data.Ord where
+
+open import Haskell.Prelude
+
+comparing : ⦃ Ord a ⦄ → (b → a) → b → b → Ordering
+comparing p x y = compare (p x) (p y)
+
+clamp : ⦃ Ord a ⦄ → (a × a) → a → a
+clamp (low , high) a = min high (max a low)


lib/Haskell/Data/List.agda
```diff
@@ -0,0 +1,110 @@
...
+
+{-----------------------------------------------------------------------------
+    Operations
+------------------------------------------------------------------------------}
+
+partition : (a → Bool) → List a → (List a × List a)
+partition p xs = (filter p xs , filter (not ∘ p) xs)
+
+private
+  elem_by : (a → a → Bool) → a → List a → Bool
```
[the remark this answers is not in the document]
HeinrichApfelmus: >  suggestion would be to put definitions not-for-export in their own nested module, which you can then open non-publicly

👍 Thanks for this tip!

> Another note is that we should avoid underscores in names because they have a special meaning in Agda (mixfix notation)

👍  Oops, sorry. I copy & pasted the definition from the source code of `base` as it was part of the definition of `nubBy`, then I rewrote that definition, and then forgot about `elem_by`.

lib/Haskell/Data/List.agda
```diff
@@ -0,0 +1,110 @@
...
+  rewrite eqSymmetry y z
+  = λ x → x
+
+-- | A deleted item is no longer an element.
+--
+prop-elem-delete
+  : ∀ ⦃ _ : Eq a ⦄ ⦃ _ : IsLawfulEq a ⦄
+      (x y : a) (zs : List a)
+  → elem x (delete y zs)
+    ≡ (if x == y then False else elem x zs)
```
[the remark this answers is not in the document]
HeinrichApfelmus: So far, I'm using a style where I try to keep most statements in the object language (Haskell) rather than in the logical metalangage (Agda) — this equality of Haskell functions is something that I would write when doing a proof about `delete` on a piece of paper. (Also, this statement covers both cases for `x == y`). Essentially, it's a single rewrite rule that concerns the computation of `elem` and discharges one `delete`.

So far, my experience has been good — this style works well with equational reasoning and is (should be) familiar to any Haskell programmer.

lib/Haskell/Data/List.agda
```diff
@@ -0,0 +1,110 @@
...
+--
+prop-elem-delete x y []
+  with x == y
+... | False = refl
+... | True  = refl
+prop-elem-delete x y (z ∷ zs)
+  with y == z in eqyz
+... | True
+    with x == z in eqxz
+    with prop-elem-delete x y zs
```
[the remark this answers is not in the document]
HeinrichApfelmus: Ah, but the induction hypothesis `recurse` has the type

```agda
prop-elem-delete x y zs
  : elem x (delete y zs)
    ≡ (if x == y then False else elem x zs)
```

and I need to rewrite `x == y` to `True` or `False` before I can use this on the right-hand side. The `using` keyword intentionally does not perform this step, I really need the `with` clause in order to get this rewriting.

lib/Haskell/Data/List.agda
```diff
@@ -0,0 +1,110 @@
...
+deleteBy : (a → a → Bool) → a → List a → List a
+deleteBy eq x ys = filter (not ∘ (eq x)) ys
+
+delete : ⦃ Eq a ⦄ → a → List a → List a
+delete = deleteBy (_==_)
+
+nubBy : (a → a → Bool) → List a → List a
+nubBy eq [] = []
+nubBy eq (x ∷ xs) = x ∷ deleteBy eq x (nubBy eq xs)
+    -- inefficient, but better for proofs
```
HeinrichApfelmus, line 31: Darn, I just noticed that when `eq` is *not* an equivalence relation, there is a chance that this definition is *not* equivalent to the definition in `Data.List`. 🙈 

For example, the Haskell report defines

```hs
nubBy            :: (a -> a -> Bool) -> [a] -> [a]
nubBy eq []      =  []
nubBy eq (x:xs)  =  x : nubBy eq (deleteBy eq xs)
```

but Agda's termination checker rejects this variant as it is not obvious that `deleteBy` does not enlarge the list. I did not think about it hard enough, but the result could be different from the definition given here for the inappropriate case where, say, `eq = _<_`. (No, don't do that. Yes, [people complain about it](https://mail.haskell.org/pipermail/libraries/2014-September/023776.html).)

I think the best course of action is to define `nubBy` as a postulate and provide another postulate 

```agda
prop-nub-cons
  : ∀ ⦃ _ : Eq a ⦄ ⦃ _ : IsLawfulEq a ⦄
      (x : a) (xs : List a)
  → nub (x ∷ xs) ≡ x ∷ delete x (nub xs)
```

that reduces `nub` for lawful instances of `Eq`, but doesn't say anything about other uses. 🤔

jespercockx: Instead of a postulate, wouldn't it be easier to just give the definition as it is in Haskell and slap a `{-# TERMINATING #-}` pragma on top? There is little doubt that this definition is actually terminating, after all. You could then prove `prop-nub-cons` instead of postulating it too.

HeinrichApfelmus: Well, the issue I perceive is that while the Haskell standard does give a definition of `nubBy`, it tacitly assumes that the predicate is an equivalence relation, and implementations do use other definitions of `nubBy`.

Specifically, the actual implementation in [base-4.21.0.0](https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-List.html#v:nubBy) is

```hs
nubBy                   :: (a -> a -> Bool) -> [a] -> [a]
#if defined(USE_REPORT_PRELUDE)
nubBy eq []             =  []
nubBy eq (x:xs)         =  x : nubBy eq (filter (\ y -> not (eq x y)) xs)
#else
-- stolen from HBC
nubBy eq l              = nubBy' l []
  where
    nubBy' [] _         = []
    nubBy' (y:ys) xs
       | elem_by eq y xs = nubBy' ys xs
       | otherwise       = y : nubBy' ys (y:xs)

-- Not exported:
-- Note that we keep the call to `eq` with arguments in the
-- same order as in the reference (prelude) implementation,
-- and that this order is different from how `elem` calls (==).
-- See #2528, #3280 and #7913.
-- 'xs' is the list of things we've seen so far,
-- 'y' is the potential new element
elem_by :: (a -> a -> Bool) -> a -> [a] -> Bool
elem_by _  _ []         =  False
elem_by eq y (x:xs)     =  x `eq` y || elem_by eq y xs
#endif
```

I don't know whether the definition that was "stolen from HBC" has the same semantics as `nubBy` in the general case — and I'm not sure that this is a problem worth solving.

Hence my inclination to add postulates for `nub` that cannot be used without `IsLawfulEq`. In this way, we can do proofs for `nub` without saying anything about the semantics of `nubBy eq` when `eq` is not an equivalence relation. Given the implementations out there, I think that's actually a good way for how people should think about `nubBy`.

----

Hm. 🤔 On the matter of postulates for `nub`, it seems to me that adding a definition of `nub` with an erased `IsLawfulEq` constraint is probably a better choice. I would also remove `nubBy` for the time being, as I don't want to invest the time into formulating a proper preconditions right now.

jespercockx: I like the approach of having `nub` be a defined function with a phantom `IsLawfulEq` constraint.

lib/Haskell/Data/List.agda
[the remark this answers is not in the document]
HeinrichApfelmus: Oh, it appears that Agda accepts the syntax

```agda
prop-elem-delete x y (z ∷ zs)
  with recurse ← prop-elem-delete x y zs
  […]
```

This will serves our gaol of naming `recurse` multiple times. Changing `with` to `using` will result in an error, you can try it to see the problem.

lib/Haskell/Data/List.agda
HeinrichApfelmus: I have removed `nubBy` and changed changed the definition of `nub` to

```agda
nub : ⦃ _ : Eq a ⦄ → @0 ⦃ IsLawfulEq a ⦄ → List a → List a
nub [] = []
nub (x ∷ xs) = x ∷ delete x (nub xs)
```

lib/Haskell/Data/List.agda
```diff
@@ -0,0 +1,103 @@
...
+------------------------------------------------------------------------------}
+
+partition : (a → Bool) → List a → (List a × List a)
+partition p xs = (filter p xs , filter (not ∘ p) xs)
+
+deleteBy : (a → a → Bool) → a → List a → List a
+deleteBy eq x ys = filter (not ∘ (eq x)) ys
+
+delete : ⦃ Eq a ⦄ → a → List a → List a
+delete = deleteBy (_==_)
```
omelkonian, line 21: As mentioned by @HeinrichApfelmus in private correspondence, this implementation deletes *all* elements instead of the *first* one as the Haskell prelude one does.

So, definitely rename this to `deleteAll` and think of adding the correct `delete` as well.

omelkonian (changes requested):

lib/Haskell/Data/List.agda
HeinrichApfelmus: I have chosen to rename this to `deleteAll` and also specialize it to `IsLawfulEq`.

I have also decided to not add `delete` at this time if that's ok with you.

lib/Haskell/Data/List.agda
```diff
@@ -0,0 +1,102 @@
...
+{-----------------------------------------------------------------------------
+    Properties
+------------------------------------------------------------------------------}
+--
+lemma-neq-trans
+  : ∀ ⦃ _ : Eq a ⦄ ⦃ _ : IsLawfulEq a ⦄
+      (x y z : a)
+  → (x == z) ≡ True
+  → (y == z) ≡ False
+  → (x == y) ≡ False
```
omelkonian, line 50: This should not be in `Data.List` as it has nothing to do with lists.

Also, it does not seem generally useful to export from anywhere, so in this case just inline the two rewrites where you use them in the proof of `prop-elem-deleteAll`.

HeinrichApfelmus: Done. Thanks, I didn't notice that I could just inline the `rewrite`.

lib/Haskell/Data/List.agda
```diff
@@ -0,0 +1,92 @@
...
+      = recurse
+...   | False
+      = recurse
+prop-elem-deleteAll x y (z ∷ zs)
+    | False
+    with x == z in eqxz
+...   | True
+      rewrite equality x z eqxz
+      rewrite eqSymmetry y z
+      rewrite eqyz
```
omelkonian, line 73: You can do this in a single `rewrite equality x z eqxz | eqSymmetry y z | eqyz`.

HeinrichApfelmus: Done. Nice, I did not realize that this single-line syntax was available.

When developing a proof, I tend to spend vertical space on `rewrite` lines, as this makes it easier to delete, add, or reorder individual rewrite steps. I started using `rewrite` as "poor man's tactics mode" or "principled man's automation", depending on your viewpoint. 😂

omelkonian: I agree that you could also write:
```agda
      rewrite equality x z eqxz
            | eqSymmetry y z
            | eqyz
```

omelkonian (approved):
LGTM

PR #397, opened 2025-01-31
Rename `Set` to `Type`

HeinrichApfelmus:
This pull request introduces the name `Type` instead of the name `Set` for all types that are transpiled to Haskell. This change has two motivations:

* Be more consistent with nomenclature from Haskell, where the canonical name for types is [`Type`](https://hackage.haskell.org/package/ghc-prim-0.13.0/docs/GHC-Types.html#t:Type).
* Avoid the name clash with [`Set`](https://hackage.haskell.org/package/containers-0.7/docs/Data-Set.html#g:1) from the `containers` package.

Using `Type` instead of `Set` removes a small, yet pervasive stumbling block for people who know Haskell and want to learn Agda2hs. In principle, people can learn the difference, but in this case, I believe that this is learning effort not well-spent.

(In a future change, it may make sense to introduce `Prop` in addition to `Type` in order to better distinguish between propositions about Haskell and types in Haskell — while every Haskell type can be transpiled from a `Type`, the converse is not true. But that discussion is out of scope for this pull request.)


test/TypeSynonyms.agda
```diff
@@ -1,5 +1,6 @@
+open import Haskell.Prim using (Type)
 
-data Nat : Set where
+data Nat : Type where
   Zero : Nat
   Suc  : Nat → Nat
 {-# COMPILE AGDA2HS Nat #-}
```
omelkonian, line 6: You missed a couple of occurrences of `Set` below.

lib/Haskell/Law/Num/Word.agda
```diff
@@ -18,7 +18,7 @@ open Num iNumWord renaming (_+_ to _+ʷ_; _*_ to _*ʷ_)
 
 postulate
   -- This is a reasonable axiom because there are more natural numbers than
-  -- integer. More precisely, Words denote a finite subset of Integers.
+  -- integer. More precisely, Words denote a finite subType of Integers.
```
omelkonian, line 21: Nope :)

omelkonian (approved):
LGTM, just a couple of minor omissions.

test/TypeSynonyms.agda
HeinrichApfelmus: Oops. Should be fixed now.

lib/Haskell/Law/Num/Word.agda
HeinrichApfelmus: Fixed.

HeinrichApfelmus:
Thanks for your review!

I have just noticed that the documentation still mentioned `Set` instead of `Type`; I have changed that as well and also added a small note for Agda users that mentions the renaming.

@omelkonian could you take another look? If you're happy, could you also merge the pull request? — I do not have sufficient permissions to merge anything, this pull request is coming from a fork.

Issue #114, opened 2022-10-06
Preserve comments


jespercockx:
As in the title, requested by @gallais

omelkonian:
Note that we can currently "fake" documentation in the generated output by including comments within a `{-# FOREIGN #-}` block.

flupe:
Linking agda/agda#5541, because if there is support for it upstream any backend would be able to preserve comments attached to definitions.

jespercockx:
This doesn't seem feasible with the current state of Agda and there is a reasonable workaround, so I'm closing this.

jespercockx:
After some discussion with @HeinrichApfelmus about the importance of being able to generate comments, I'm re-opening this.

In particular, the existence of this tool shows that there is a real need: https://github.com/cardano-foundation/cardano-wallet-agda/tree/main/lib/agda2hs-fixer If necessary, we could consider re-parsing the actual Agda source in agda2hs just so we can properly extract the comments.

PR #406, opened 2025-04-11
Add `containers.agda-lib` to `lib` directory

HeinrichApfelmus:
This pull request adds a library `lib/containers/containers.agda-lib` which imports large chunks of the `Data.Map` and `Data.Set` modules from the [containers](https://hackage.haskell.org/package/containers) package, including proofs.

### Comments

* I have chosen to not add the prefix `Haskell.` to the module names `Data.Map.Prop` and `Data.Set.Prop`, because:

    * Strictly speaking, these modules do not correspond to an existing Haskell module, …
    * … but the properties proven have value as exported Haddock comments. In other words, once we can export properties to comments (related to #114), we can later make `Data.Map.Prop` part of, say, a `containers-prop.cabal` package and make the proven properties available as a Haskell package. (The properties cannot be reused for proofs in Haskell Land, but at least they can be exported as "this has been proven!" to Haskell Land.)

* I was not quite sure whether to choose the name `Data.Set.Prop` or `Data.Set.Law`. [John Hughes](https://belle.sourceforge.net/doc/hughes95design.pdf) appears to use both "law" and "property" synonymously, while using "law" more often. Perhaps a good distinction is that the "laws" are closer to being complete while "properties" are mostly "haphazard" consequences.

### DIFF
--- a/lib/containers/agda/EverythingContainers.agda
+++ b/lib/containers/agda/EverythingContainers.agda
@@ -1,4 +1,0 @@
-module EverythingContainers where
-
-import Data.Map
-import Data.Set
--- a/lib/containers/agda/EverythingContainers.agda
+++ b/lib/containers/agda/EverythingContainers.agda
@@ -1,0 +1,4 @@
+module EverythingContainers where
+
+import Data.Map
+import Data.Set
--- a/lib/containers/agda/Haskell/Data/Map.agda
+++ b/lib/containers/agda/Haskell/Data/Map.agda
@@ -1,4 +1,3 @@
-{-# OPTIONS --erasure #-}
 
 -- | Postulates and definitions of the operations supported by 'Map'.
 module Haskell.Data.Map where
@@ -13,13 +12,15 @@ open import Haskell.Data.Maybe using
 
 import Data.Map.Maybe as Maybe
 import Haskell.Prelude as List using (map)
+
+open import Data.Set using (Set)
 import Data.Set as Set
 
 {-----------------------------------------------------------------------------
     Helper predicates
 ------------------------------------------------------------------------------}
-AntitonicPred : {a : Set} → {{Ord a}} → (a → Bool) → Set
-AntitonicPred {a} p =
+Antitonic : {a : Type} → {{Ord a}} → (a → Bool) → Type
+Antitonic {a} p =
   ∀ {x y : a} → ((x <= y) ≡ True) → ((p x >= p y) ≡ True)
 
 {-----------------------------------------------------------------------------
@@ -27,9 +28,9 @@ AntitonicPred {a} p =
     involving 1 value type
 ------------------------------------------------------------------------------}
 postulate
-  Map : ∀ (k : Set) → Set → Set
+  Map : ∀ (k : Type) → Type → Type
 
-module _ {k a : Set} {{_ : Ord k}} where
+module _ {k a : Type} {{_ : Ord k}} where
   postulate
     lookup    : k → Map k a → Maybe a
     null      : Map k a → Bool
@@ -44,7 +45,7 @@ module _ {k a : Set} {{_ : Ord k}} where
   keys : Map k a → List k
   keys = List.map fst ∘ toAscList
 
-  keysSet : Map k a → Set.ℙ k
+  keysSet : Map k a → Set k
   keysSet = Set.fromList ∘ keys
 
   size : Map k a → Int
@@ -120,13 +121,13 @@ module _ {k a : Set} {{_ : Ord k}} where
         ≡ Maybe.filter (p key) (lookup key m)
 
     prop-lookup-takeWhileAntitone
-      : ∀ (p : k → Bool) → AntitonicPred p
+      : ∀ (p : k → Bool) → Antitonic p
       → ∀ (key : k) (m : Map k a)
       → lookup key (takeWhileAntitone p m)
         ≡ lookup key (filterWithKey (λ k _ → p k) m)
 
     prop-lookup-dropWhileAntitone
-      : ∀ (p : k → Bool) → AntitonicPred p
+      : ∀ (p : k → Bool) → Antitonic p
       → ∀ (key : k) (m : Map k a)
       → lookup key (dropWhileAntitone p m)
         ≡ lookup key (filterWithKey (λ k _ → not (p k)) m)
@@ -145,10 +146,10 @@ module _ {k a : Set} {{_ : Ord k}} where
     Nothing → insert k new m
     (Just old) → insert k (f new old) m
 
-  withoutKeys : Map k a → Set.ℙ k → Map k a
+  withoutKeys : Map k a → Set k → Map k a
   withoutKeys m s = filterWithKey (λ k _ → not (Set.member k s)) m
 
-  restrictKeys : Map k a → Set.ℙ k → Map k a
+  restrictKeys : Map k a → Set k → Map k a
   restrictKeys m s = filterWithKey (λ k _ → Set.member k s) m
 
   filter : (a → Bool) → Map k a → Map k a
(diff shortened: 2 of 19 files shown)


lib/containers/agda/EverythingContainers.agda
```diff
@@ -0,0 +1,4 @@
+module EverythingContainers where
```
omelkonian, line 1: I would prefer giving Everything files the same name as the library, i.e. name this `containers.agda`.

lib/containers/containers.agda-lib
```diff
@@ -0,0 +1,5 @@
+name: containers
+include: agda
+depend:
+  agda2hs
+flags: -W noUnsupportedIndexedMatch --erasure
```
omelkonian, line 5: Do you really need `-W noUnsupportedIndexedMatch`?

lib/containers/agda/Haskell/Data/Set.agda
```diff
@@ -0,0 +1,130 @@
...
+
+open import Haskell.Prelude hiding (lookup; null; map; filter)
+
+{-----------------------------------------------------------------------------
+    Data.Set
+------------------------------------------------------------------------------}
+
+-- | As the term 'Set' is already taken in Agda, we use ℙ (\bP).
+postulate
+  ℙ : Set → Set
```
omelkonian, line 14: Nasty clash with powerset notation; why not require the `--no-import-sorts` flag and use `Set` as expected?

lib/containers/agda/Haskell/Data/Map.agda
```diff
@@ -0,0 +1,230 @@
...
+{-----------------------------------------------------------------------------
+    Operations
+    involving 2 value types
+------------------------------------------------------------------------------}
+module _ {k a b : Set} {{_ : Ord k}} where
+  postulate
+
+    instance
+      iMapFunctor : Functor (Map k)
+
```
omelkonian, line 184: This shouldn't be parametrized by `a` and `b`, so move outside the module?

lib/containers/agda/Haskell/Data/Map.agda
```diff
@@ -0,0 +1,230 @@
...
+    )
+
+import Data.Map.Maybe as Maybe
+import Haskell.Prelude as List using (map)
+import Data.Set as Set
+
+{-----------------------------------------------------------------------------
+    Helper predicates
+------------------------------------------------------------------------------}
+AntitonicPred : {a : Set} → {{Ord a}} → (a → Bool) → Set
```
omelkonian, line 21: Rename to simply `Antitonic`.

lib/containers/agda/Haskell/Data/Map.agda
```diff
@@ -0,0 +1,230 @@
...
+  elems = List.map snd ∘ toAscList
+
+  keys : Map k a → List k
+  keys = List.map fst ∘ toAscList
+
+  keysSet : Map k a → Set.ℙ k
+  keysSet = Set.fromList ∘ keys
+
+  size : Map k a → Int
+  size = length ∘ toAscList
```
omelkonian, line 51: Is it crucial for computation that you define this naive versions, or would it be better to postulate these as well?

omelkonian (changes requested):
Great stuff!

What is urgently missing before we can merge is a test suite that ensures we get the desired Haskell output.

Other than that, you've imposed a rather idiosyncratic style throughout that should be reversed in favour of consistency with the rest of the repo. It'd be easier if I pushed a styling commit after yours if you gave me access to your branch. Otherwise, let me know if you need explicit indications.`

lib/containers/containers.agda-lib
HeinrichApfelmus: As I have no idea what this extension is doing — probably not. 😂 This was an instance of cargo-cult copying.

lib/containers/agda/Haskell/Data/Map.agda
HeinrichApfelmus: Having `member` as computation is useful, because we can avoid to manually rewriting `member` to `lookup` in this way, which is something that we almost always want to do.

For the other three, the computation is mildly useful — the definition of `length` is powerful enough so that this computation does become useful. The effect is that all these functions rewrite to properties of `toAscList`. Unfortunately, this module does not include a complete axiomatization of `toAscList` — I would like to add that at a later point, not in this pull request. So, for these functions I don't mind either way — we probably need to change it later anyway.

I like the following viewpoint:

* computation = equation + rewrite rule.

Then, the question becomes whether the we want the rewrite rule (→ write it as a definition) or not (→ postulate an equation).

In this light, there is a general case to be made for the axioms of `lookup` to become rewrite rules — but that's definitely not a question for this pull request.

HeinrichApfelmus:
Thanks! 😊 

> What is urgently missing before we can merge is a test suite that ensures we get the desired Haskell output.

As this code binds an existing Haskell library, it does not produce any Haskell output by itself. I suppose the next best thing is that we add an Agda module that imports `Data.Map` and has non-trivial Haskell output, so we check that `Data.Map` is imported correctly?

Or do you mean the "Haddock output" for e.g. `Data.Map.Prop` that I have alluded to? That would require more support from `agda2hs`, e.g. by moving functionality from [agda2hs-fixer](https://github.com/cardano-foundation/cardano-wallet-agda/tree/cfcf563f3455d104bda373c4416518d4fb342e95/lib/agda2hs-fixer) to `agda2hs`.

At the very least, we can check in CI that the proofs in `Data.Map.Prop` and `Data.Set.Prop` are correct!

> Other than that, you've imposed a rather idiosyncratic style throughout that should be reversed in favour of consistency with the rest of the repo. It'd be easier if I pushed a styling commit after yours if you gave me access to your branch. Otherwise, let me know if you need explicit indications.`

Yeah, sorry about that. 😅 I like to squander vertical space for the purpose of Haddock output. I also like `rewrite` — it saves a lot of time as opposed to explicit `cong` and equational reasoning.

This branch is coming from a fork, I have invited you to my fork. (Alternatively, you could invite me to the original repo, I don't mind either way).

omelkonian:
> Yeah, sorry about that. 😅 I like to squander vertical space for the purpose of Haddock output. I also like rewrite — it saves a lot of time as opposed to explicit cong and equational reasoning.

OK, you have plans for producing Haddock, so that makes sense now. Then it's OK to leave the style as is for now, and revisit later if needed.

Perfectly fine to use `rewrite`, just make sure you group together with `rewrite H1 | H2 | ... | Hn`.

omelkonian:
> As this code binds an existing Haskell library, it does not produce any Haskell output by itself. I suppose the next best thing is that we add an Agda module that imports Data.Map and has non-trivial Haskell output, so we check that Data.Map is imported correctly?

Exactly! A simple test that uses a handful of operations on sets and maps and typechecks correctly with proper imports.

> Or do you mean the "Haddock output" for e.g. Data.Map.Prop that I have alluded to? That would require more support from agda2hs, e.g. by moving functionality from [agda2hs-fixer](https://github.com/cardano-foundation/cardano-wallet-agda/tree/cfcf563f3455d104bda373c4416518d4fb342e95/lib/agda2hs-fixer) to agda2hs.

Nah, it's fine to leave this for the future.

HeinrichApfelmus:
> Exactly! A simple test that uses a handful of operations on sets and maps and typechecks correctly with proper imports.

I have now added a commit with two modules `Test.Agda2Hs.Data.Map` and `Test.Agda2Hs.Data.Set` that do this.

I have decided to put them in a `containers-prop.cabal` cabal package, because:

* When exporting `Data.Map.Prop` as Haddock in the future, the plan is to do it as part of a package `containers-prop`.
* This setup keeps the CI simple — `cabal` takes the place of `make` rules.

omelkonian (approved):
Excellent!!

PR #407, opened 2025-05-02
Add `containers-lib` to nix flake `packages`

HeinrichApfelmus:
This pull request adds an attribute `containers-lib` to the `packages` output in `flake.nix`. In this way, users can provision the `containers-lib` using Nix.

This is an addendum to #406 to make `containers.agda-lib` useable with Nix.

Note: This pull request also changes the version of the `base-lib` attribute to match the version `4.18` of the `base` package on Hackage.

### DIFF
--- a/.github/workflows/nix-ci.yml
+++ b/.github/workflows/nix-ci.yml
@@ -24,12 +24,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        derivation: [agda2hs, base-lib]
+        derivation: [agda2hs, base-lib, containers-lib]
         include:
           - pretty: "Compile agda2hs"
             derivation: agda2hs
           - pretty: "Typecheck with Agda"
             derivation: base-lib
+          - pretty: "Typecheck with Agda"
+            derivation: containers-lib
     steps:
     - uses: actions/checkout@v3
     - uses: cachix/install-nix-action@v22
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -5,10 +5,11 @@
 let
   lib = import ./lib.nix { inherit pkgs; };
   version = "1.3";
+
   base-lib = pkgs.agdaPackages.mkDerivation {
     pname = "base";
     meta = { };
-    version = version;
+    version = "4.18";
     preBuild = ''
       echo "{-# OPTIONS --sized-types #-}" > Everything.agda
       echo "module Everything where" >> Everything.agda
@@ -19,5 +20,5 @@ let
 in
 {
   inherit (lib) agda2hs;
-  base-lib = base-lib;
+  inherit base-lib;
 }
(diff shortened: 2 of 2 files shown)


jespercockx (approved):

HeinrichApfelmus:
@jespercockx Thanks! Could you merge the pull request? — I don't have any permissions on this repository. 😅

HeinrichApfelmus:
Actually, I think that there is something wrong when Nix typechecks the `containers.agda-lib` library: It tries to write local interface files for `base.agda-lib` into the nix store.

Somehow, transitive dependencies do not compose just yet (`containers` depends on `base`), not sure what to do here.

liesnikov:
do you mean these lines? https://github.com/agda/agda2hs/actions/runs/14793443918/job/41535343556#step:5:204 (and till 304 or so)
Because it seems like that was just a dependency build, so a separate derivation, as it should be. And then when containers it typechecked I don't see any re-checking of base.

HeinrichApfelmus:
> do you mean these lines? https://github.com/agda/agda2hs/actions/runs/14793443918/job/41535343556#step:5:204 (and till 304 or so)

Yes and no.

> And then when containers it typechecked I don't see any re-checking of base.

Yes. However, the problem is that `agdaPackages.mkDerivation` produces a result that is incompatible with `agda2hs` when used. I.e. `agdaPackages` produces self-consistent results in the nix store, but those results are not useable when trying to compile with `adga2hs`. 

We need an equivalent `agda2hs.mkDerivation`. I have some code here:

https://github.com/HeinrichApfelmus/bags/blob/main/nix/lib.nix

but haven't managed to backport it yet.

liesnikov:
Hmm, I'm not sure I understand exactly. What do you consider to be a "built derivation" for agda2hs?

If it's agda+agdai files and agda2hs then doesn't behave the same way agda proper does, I think I see the problem. I'd be surprised if this is the case.
Or is it agda, agdai, and hs files, i.e. a self-contained compiled unit for all usages of agda2hs?

HeinrichApfelmus:
> If it's agda+agdai files and agda2hs then doesn't behave the same way agda proper does, I think I see the problem. I'd be surprised if agda2hs behaved very differently here, but it can be.

Yep, that's indeed the case and the problem.

* `Agda` puts `.agdai` files in a directory `_build/2.7.0`
* The nixified `agda2hs` uses `--local-interfaces` (here: https://github.com/agda/agda2hs/blob/256e604c4f7b1951459efdf68a7184e83e1f0d2a/nix/agda2hs.nix#L26 ) and expects the `.agdai` files to be colocated with the `.agda` files.

Consequently, when compiling a downstream file with `agda2hs`, the `--local-interfaces` options causes `agda2hs` to attempt to write an `.agdai` file to the nix store.

liesnikov:
Ah, indeed.
That file should probably stay in sync with nixpkgs' source for agda. I just couldn't find a way to change the (ab)use the upstream in a way that would work reliably, but it doesn't mean that there isn't one.

PR #410, opened 2025-05-12
switch from magic-nix-cache to nix-community/cache-nix-action

liesnikov:
magic nix cache that was used in the nix ci is discontinued since February 2025 due to Github switching to a new cache service
nix-community/cache-nix-action seems to do the job for now, but there's probably some tweaking that can be done to reduce the cache sizes or sharing

https://determinate.systems/posts/magic-nix-cache-free-tier-eol/
https://github.blog/changelog/2025-03-20-notification-of-upcoming-breaking-changes-in-github-actions/#decommissioned-cache-service-brownouts

### DIFF
--- a/.github/workflows/nix-ci.yml
+++ b/.github/workflows/nix-ci.yml
@@ -32,8 +32,16 @@ jobs:
             derivation: base-lib
     steps:
     - uses: actions/checkout@v3
-    - uses: cachix/install-nix-action@v22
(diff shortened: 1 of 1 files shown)


jespercockx:
Thanks for the PR!

liesnikov:
hmm it looks like magic-nix-cache is alive again

https://determinate.systems/posts/bringing-back-magic-nix-cache-action/

Issue #411, opened 2025-05-22
Monad, Applicative, Functor: Add equations for default definitions and superclass methods


HeinrichApfelmus:
When working with the `Monad`, `Applicative` and `Functor` classes, I found some room for improvement.

Specifically, I have formalized a [proof](https://github.com/HeinrichApfelmus/bags/blob/57ff6987e48b16f6236bea030d63163299921a22/lib/bags/agda/Control/Monad/Prop.agda#L78-L84) that the `Monad` laws imply the `Applicative` laws which I would like to backport to `base.agda-lib`.

Currently, the `Monad` class is split into the full class, and the default implementation for some methods:

```agda
  record Monad (m : Type → Type) : Type₁ where
    field
      _>>=_ : m a → (a → m b) → m b
      overlap ⦃ super ⦄ : Applicative m
      return : a → m a
      _>>_ : m a → (@0 {{ a }} → m b) → m b
  -- ** defaults
  record DefaultMonad (m : Type → Type) : Type₁ where
    field
      _>>=_ : m a → (a → m b) → m b
      overlap ⦃ super ⦄ : Applicative m
    return : a → m a
    return = pure

    _>>_ : m a → (@0 {{ a }} → m b) → m b
    m >> m₁ = m >>= λ x → m₁ {{x}}
```

I would like to argue that:

1. The class `Monad` should include equations for all default methods, such as

    ```agda
      prop->>->>=
        : ∀ ⦃ _ : Monad m ⦄ ⦃ _ : IsLawfulMonad m ⦄
        → (mx : m a) (my : @0 ⦃ a ⦄ → m b)
        → mx >> my  ≡  mx >>= (λ x → my ⦃ x ⦄)
    ```

    In other words, we can rely on the fact that `(>>)` is propositionally equal to its default definition, we just cannot assume that this equality holds by computation.

    I think it's fair to put these equalities in the `Monad` class as opposed to the `IsLawfulMonad` class, because these are about default definitions of functions, not about properties that these functions should have.
  
    By the way, as far as I can tell, with the current setup of `IsLawfulMonad`, I don't think that it's possible to prove `prop->>->>=` — I had to `postulate` it. :O

2. Similarly, the class `Monad` should include equations for all methods from superclasses, such as the definition of `fmap` in terms of `>>=` and `return`:

    ```agda
      prop-fmap->>=
        : ∀ {a b} (f : a → b) (ma : m a)
        → fmap f ma ≡ (ma >>= (return ∘ f))    
[...]

PR #412, opened 2025-05-24
Reorganize Monad laws, prove Applicative laws

HeinrichApfelmus:
This pull request reorganizes the `IsLawfulMonad` class.

The main reason for this reorganization is that `IsLawfulFunctor` and `IsLawfulApplicative` are actually consequences of `MonadLaws` and `IsDefaultMonad`. However, this fact is difficult to use in the old definition as the other two classes are used as superclasses.

Context: #411

### Comments

While not in the scope of this pull request, possible future reorganizations are:

1. Drop the `IsLawfulApplicative` superclass from `IsLawfulMonad` and make it a global consequence instead, i.e. add a global `instance _ : {{IsLawfulMonad m}} → IsLawfulApplicative m`.

    This has the slight drawback that the existing instance of `IsLawfulApplicative` need to be removed while making sure that `Haskell.Law.Applicative` re-exports the global instance. Then again, the `IsLawfulMonad` instance is the easiest way to prove `IsLawfulApplicative`, and in the spirit of proof-irrelevance, we don't actually care how we have proven `IsLawfulMonad`.

2. Merge the `IsDefaultMonad` class into the `Monad` class.

    Reason: In Haskell, it is traditional to expect the `Monad` constraint to imply the laws mentioned in the documentation. For usage sites, this means that `Monad` is sufficient, the `IsLawfulMonad` constraint is conceptually implied by the `Monad` class. This conceptual model works until we actually have to formalize proofs, at which point we want to develop proofs incrementally. For incremental development of proofs, the separation between "implementation" (`Monad`) and "properties" (`IsLawfulMonad`) is very useful, but I would argue that this case rarely happens specifically for the `Monad` class, where instances are seldomly defined by library users. (In contrast, `Eq` and `Ord` are frequently defined by library users for their own data types, and the separation of `IsLawfulEq` and `IsLawfulOrd` is more valueable.)

    As a first step towards conceptual unification, I would propose to merge `IsDefaultMonad` into the `Monad` class, as the equality to the default implementations is uncontroversial.

5. Merge the `IsLawfulMonad` class into the `Monad` class. Same reasoning as in 2., but more extreme.

### DIFF
--- a/lib/base/Haskell/Law/Monad/Def.agda
+++ b/lib/base/Haskell/Law/Monad/Def.agda
@@ -9,34 +9,63 @@ open import Haskell.Prim.Monoid
 open import Haskell.Prim.Tuple
 
 open import Haskell.Law.Applicative
+open import Haskell.Law.Equality
+open import Haskell.Law.Extensionality
 
-record IsLawfulMonad (F : Type → Type) ⦃ iMonadF : Monad F ⦄ : Type₁ where
+-- Helper function: Substitution in the second argument of '(>>=)'.
+cong-monad
+  : ∀ ⦃ _ : Monad m ⦄ (mx : m a) {f g : a → m b}
+  → (∀ x → f x ≡ g x)
+  → (do x ← mx; f x) ≡ (do x ← mx; g x)
+--
+cong-monad mx {f} {g} eq = cong (mx >>=_) (ext eq)
+
+-------------------------------------------------------------------------------
+-- Monad laws
+--
+-- `PreLawfulMonad` contains all laws that we expect a 'Monad' to satisfy,
+-- except that we do not yet require that the superclasses are themselves
+-- lawful, this is deferred to `IsLawfulMonad`.
+-- The lawfulness of the superclasses can be proven from `PreLawfulMonad`.
+
+record PreLawfulMonad (m : Type → Type) ⦃ _ : Monad m ⦄ : Type₁ where
   field
-    overlap ⦃ super ⦄ : IsLawfulApplicative F
+    -- The three monad laws
+    leftIdentity  : ∀ {a} (x : a) (k : a → m b)
+      → (return x >>= k) ≡ k x
 
-    -- Left identity: return a >>= k = k a
-    leftIdentity : {a : Type} → (a' : a) (k : a → F b) → ((return a') >>= k) ≡ k a'
+    rightIdentity : ∀ {a} (ma : m a)
+      → (ma >>= return) ≡ ma
 
-    -- Right identity: m >>= return = m
-    rightIdentity : {a : Type} → (ma : F a) → (ma >>= return) ≡ ma
-
-    -- Associativity: m >>= (\x -> k x >>= h) = (m >>= k) >>= h
-    associativity : {a b c : Type} → (ma : F a) (f : a → F b) (g : b → F c)
+    associativity : ∀ {a b c} (ma : m a) (f : a → m b) (g : b → m c)
       → (ma >>= (λ x → f x >>= g)) ≡ ((ma >>= f) >>= g)
 
-    -- pure = return
-    pureIsReturn : (a' : a) → pure a' ≡ (Monad.return iMonadF a')
-    -- m1 <*> m2 = m1 >>= (\x1 -> m2 >>= (\x2 -> return (x1 x2)))
-    sequence2bind : {a b : Type} → (mab : F (a → b)) (ma : F a)
-      → (mab <*> ma) ≡ (mab >>= (λ x1 → (ma >>= (λ x2 → return (x1 x2)))))
+    -- Default functions
+    def->>->>= : ∀ {a b} (ma : m a) (mb : m b)
+      → ma >> mb ≡ ma >>= (λ x → mb)
 
-    -- fmap f xs  =  xs >>= return . f
-    fmap2bind : {a b : Type} → (f : a → b) (ma : F a)
-      → fmap f ma ≡ (ma >>= (return ∘ f))
-    -- (>>) = (*>)
-    rSequence2rBind : (ma : F a) → (mb : F b) → (ma *> mb) ≡ (ma >> mb)
+    def-pure-return : ∀ {a} (x : a)
+      → pure {m} x ≡ return x
+
+    -- Superclass functions
+    def-fmap->>= : ∀ {a b} (f : a → b) (ma : m a)
+      → fmap f ma ≡ ma >>= (return ∘ f)
+
+    def-<*>->>= : ∀ {a b} (mab : m (a → b)) (ma : m a)
+      → (mab <*> ma) ≡ (mab >>= (λ f → (ma >>= (λ x → return (f x)))))
+
+open PreLawfulMonad ⦃ ... ⦄ public
+
+-- All laws together
+record IsLawfulMonad (m : Type → Type) ⦃ _ : Monad m ⦄ : Type₁ where
+  field
+    overlap ⦃ applicative ⦄ : IsLawfulApplicative m
+    overlap ⦃ monad ⦄       : PreLawfulMonad m
 
 open IsLawfulMonad ⦃ ... ⦄ public
+
+-------------------------------------------------------------------------------
+-- postulated monad laws, to be proven
 
 instance postulate
   iLawfulMonadFun : IsLawfulMonad (λ b → a → b)
--- a/lib/base/Haskell/Law/Monad/Maybe.agda
+++ b/lib/base/Haskell/Law/Monad/Maybe.agda
@@ -10,26 +10,26 @@ open import Haskell.Law.Monad.Def
 open import Haskell.Law.Applicative.Maybe
 
 instance
-  iLawfulMonadMaybe : IsLawfulMonad Maybe
-  iLawfulMonadMaybe .leftIdentity _ _ = refl
+  iPreLawfulMonadMaybe : PreLawfulMonad Maybe
+  iPreLawfulMonadMaybe = λ where
+    .leftIdentity _ _ → refl
+    .rightIdentity → λ where
+      Nothing  → refl
+      (Just _) → refl
+    .associativity → λ where
+      Nothing  _ _ → refl
+      (Just _) _ _ → refl
 
-  iLawfulMonadMaybe .rightIdentity = λ { Nothing → refl; (Just _) → refl }
+    .def->>->>= _ _ → refl
+    .def-pure-return _ → refl
 
-  iLawfulMonadMaybe .associativity = λ { Nothing _ _ → refl; (Just _) _ _ → refl }
+    .def-fmap->>= _ → λ where
+      Nothing → refl
+      (Just _) → refl
+    .def-<*>->>= → λ where
+      Nothing  _        → refl
+      (Just _) Nothing  → refl
+      (Just _) (Just _) → refl
 
-  iLawfulMonadMaybe .pureIsReturn _ = refl
-
-  iLawfulMonadMaybe .sequence2bind =
-    λ { Nothing  _        → refl
-      ; (Just _) Nothing  → refl
-      ; (Just _) (Just _) → refl
-      }
-
-  iLawfulMonadMaybe .fmap2bind = λ { _ Nothing → refl; _ (Just _) → refl }
-
-  iLawfulMonadMaybe .rSequence2rBind =
-    λ { Nothing  _        → refl
-      ; (Just _) Nothing  → refl
-      ; (Just _) (Just _) → refl
-      }
-
+  iIsLawfulMonadMaybe : IsLawfulMonad Maybe
+  iIsLawfulMonadMaybe = record {}
--- a/lib/base/Haskell/Law/Applicative/FromMonad.agda
+++ b/lib/base/Haskell/Law/Applicative/FromMonad.agda
@@ -1,0 +1,129 @@
+module Haskell.Law.Applicative.FromMonad where
+
+open import Haskell.Prim
+
+open import Haskell.Prim.Applicative
+open import Haskell.Prim.Functor
+open import Haskell.Prim.Monad
+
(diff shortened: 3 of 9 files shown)


jespercockx (approved):

jespercockx:
Regarding incremental development: ideally there should be a gradual path:

1. An implementation without any specification.
2. An implementation that states its specification but doesn't prove it.
3. An implementation together with a specification and a proof.

The current design allows all three: one can just implement a `Monad` instance at first, then later add a postulate that it is lawful (which can already be used in other proofs that use the monad) and finally replace the postulate with an actual proof. If we merge `IsLawfulMonad` into the `Monad` class then it is no longer possible to have *just* an implementation of the monad interface without making any claims about its behavior.

Another argument against merging the laws into the class is that sometimes, Haskell programmers actually *do* implement non-lawful monads, and then it would be inconsistent to postulate that such a monad is lawful.

jespercockx:
It should be possible to add quotient types (with postulates + perhaps a rewrite rule) so I'd say that's not a showstopper for this PR. But since it's not really connected to this issue, please make a separate issue for that.

I do agree that not needing a separate constraint for the laws is nice. But this could in principle also be achieved by defining a bundled class `LawfulMonad` that contains both a monad and its IsLawful proof (and has eta-equality).

lib/base/Haskell/Law/Monad/Def.agda
```diff
@@ -9,35 +9,70 @@ open import Haskell.Prim.Monoid
...
+      → (mab <*> ma) ≡ (mab >>= (λ f → (ma >>= (λ x → return (f x)))))
+
+open IsDefaultMonad ⦃ ... ⦄ public
+
+-- All laws together
+record IsLawfulMonad (m : Type → Type) ⦃ iMonadm : Monad m ⦄ : Type₁ where
+  field
+    overlap ⦃ applicative ⦄ : IsLawfulApplicative m
+    overlap ⦃ default ⦄     : IsDefaultMonad m
+    overlap ⦃ monad ⦄       : MonadLaws m
```
omelkonian, line 69: While I agree with the MonadLaws/IsLawfulMonad separation, I don't think we should also separate IsDefaultMonad/MonadLaws; one only has the monad laws proven when the user's instance has implemented correctly the default fields.

lib/base/Haskell/Law/Monad/Def.agda
```diff
@@ -9,35 +9,70 @@ open import Haskell.Prim.Monoid
...
+
+-- Default and superclass functions
+-- are defined in terms of '>>=' and 'return'
+record IsDefaultMonad (m : Type → Type) ⦃ iMonad : Monad m ⦄ : Type₁ where
+  field
+    def->>->>= : ∀ {a b} (ma : m a) (mb : m b)
+      → ma >> mb ≡ ma >>= (λ x → mb)
+
+    def-pure-return : ∀ {a} (x : a)
+      → pure x ≡ return {{iMonad}} x
```
omelkonian, line 55: The problem here is code repetition, i.e. the right hand sides are simply inlining the DefaultMonad definitions.

Here's a suggested solution:

- 1) make Monad/DefaultMonad more modular
```agda
module Do where

  -- ** base
  record Monad (m : Type → Type) : Type₁ where
    field
      _>>=_ : m a → (a → m b) → m b
      overlap ⦃ super ⦄ : Applicative m
      return : a → m a
      _>>_ : m a → (@0 {{ a }} → m b) → m b

  -- ** defaults
  module MonadDefaults (m : Type → Type) where
    >>-from->>= : (m a → (a → m b) → m b) → (m a → (@0 ⦃ a ⦄ → m b) → m b)
    >>-from->>= _>>=_ = λ m m₁ →
      m >>= λ x → m₁ {{x}}

    return-from-pure : (a → m a) → (a → m a)
    return-from-pure pure = pure

  record DefaultMonad (m : Type → Type) : Type₁ where
    field
      _>>=_ : m a → (a → m b) → m b
      overlap ⦃ super ⦄ : Applicative m

    open MonadDefaults m

    return : a → m a
    return = return-from-pure pure

    _>>_ : m a → (@0 {{ a }} → m b) → m b
    _>>_ = >>-from->>= _>>=_
```

- 2) restate conformance laws to default implementations 
```agda
record MonadLaws ... where
  -- monad laws
  ...
  
  -- monad-functor laws
  ...

  -- monad-applicative laws
  ...

  -- proper default implementations
  private module def = MonadDefaults m

  field
    def->>->>= : ∀ {a b} (ma : m a) (mb : m b)
      → ma >> mb ≡ def.>>-from->>= _>>=_ ma mb

    def-pure-return : ∀ {a} (x : a)
      → return x ≡ def.return-from-pure pure x
```

Note that the whole construction can be automated with reflection (e.g. automatically get the conformance laws given the default `record`s), as are the boring pointwise unfoldings of the equations (i.e. define an "n-ary" relation to check conformance of any two fields).

lib/base/Haskell/Law/Monad/Maybe.agda
```diff
@@ -10,26 +10,21 @@ open import Haskell.Law.Monad.Def
...
+  iMonadLawsMaybe .associativity (Just x) = λ f g → refl
+
+  iIsDefaultMonadMaybe : IsDefaultMonad Maybe
+  iIsDefaultMonadMaybe .def->>->>= = λ ma mb → refl
+  iIsDefaultMonadMaybe .def-pure-return = λ x → refl
+  iIsDefaultMonadMaybe .def-fmap->>= f Nothing  = refl
+  iIsDefaultMonadMaybe .def-fmap->>= f (Just x) = refl
+  iIsDefaultMonadMaybe .def-<*>->>= Nothing = λ ma → refl
+  iIsDefaultMonadMaybe .def-<*>->>= (Just x) Nothing = refl
+  iIsDefaultMonadMaybe .def-<*>->>= (Just x) (Just y) = refl
```
omelkonian, line 27: Please write a less verbose instance (closer to the previous one), which does not repeat the names `iMonadLawsMaybe` and `iIsDefaultMonadMaybe` unnecessarily.

omelkonian: Combined with the other comments I made, I'd expect the instance to look something like this:
```agda
instance
  iMonadLawsMaybe : MonadLaws Maybe
  iMonadLawsMaybe = λ where
    .leftIdentity _ _ → refl
    .rightIdentity → λ where
      Nothing  → refl
      (Just _) → refl
    .associativity → λ where
      Nothing  _ _ → refl
      (Just _) _ _ → refl
    .fmap->>= _ → λ where
      Nothing  → refl
      (Just _) → refl
    .<*>->>= → λ where
      Nothing  _        → refl
      (Just _) Nothing  → refl
      (Just _) (Just _) → refl
    .def->>->>= _ _ → refl
    .def-pure-return _ → refl
```

lib/base/Haskell/Law/Monad/Def.agda
```diff
@@ -9,35 +9,70 @@ open import Haskell.Prim.Monoid
...
+    def-fmap->>= : ∀ {a b} (f : a → b) (ma : m a)
+      → fmap f ma ≡ ma >>= (return ∘ f)
+
+    def-<*>->>= : ∀ {a b} (mab : m (a → b)) (ma : m a)
+      → (mab <*> ma) ≡ (mab >>= (λ f → (ma >>= (λ x → return (f x)))))
+
+open IsDefaultMonad ⦃ ... ⦄ public
+
+-- All laws together
+record IsLawfulMonad (m : Type → Type) ⦃ iMonadm : Monad m ⦄ : Type₁ where
```
omelkonian, line 65: Unnecessary `iMonadm`.

omelkonian (changes requested):
Looks good overall, I've just pointed some design changes to make the approach more maintanable.

omelkonian:
1. I agree with @jespercockx in keeping `Monad` and `IsLawfulMonad` separate.

> HEINRICH: Just for the record: The benefit of merging the two classes would be that any general proposition about monads has a shorter type, because it only needs to mention ⦃ _ : Monad m ⦄ as an argument, as opposed to the longer ⦃ _ : Monad m ⦄ ⦃ _ : IsLawfulMonad m ⦄.
> JESPER: I do agree that not needing a separate constraint for the laws is nice. But this could in principle also be achieved by defining a bundled class LawfulMonad that contains both a monad and its IsLawful proof (and has eta-equality).

Yep, that's not a valid argument for merging. One could also make this construction more [generic](https://omelkonian.github.io/formal-prelude/Prelude.ConjClass.html) so that it works for any "typeclass conjuction" (although it breaks once you move to higher universe levels..., but since we only use `Type`s at level 0 we should be fine here).

lib/base/Haskell/Law/Monad/Def.agda
HeinrichApfelmus: The nomenclature "monad laws" typically refers to the three identities I originally put in the record `MonadLaws` — and only those three identities.

For example, Hutton, "Programming in Haskell", (2018), p. 174 writes:

> Monad laws
>
> In a similar manner to functors and applicatives, the two monadic primitives are required to satisfy some equational laws:
>
>     return x >>= f    =  f x
>     mx >>= return     =  mx
>     (mx >>= f) >>= g  =  mx >>= (\ x -> (f x >>= g))

The identities that I had put in the `IsDefaultMonad` record concern the definition of functions that are traditional or relate to superclasses, but are not part of the "monad laws", such as `pure`, `>>`, `fmap`, and `(<*>)`.

In other words, in principle, it is possible for a user instance of `Monad` to implement the "monad laws" correctly while implementing the traditional and superclass functions incorrectly.

(The documentation for [`Control.Monad`](https://hackage.haskell.org/package/base-4.21.0.0/docs/Control-Monad.html#t:Monad) in the `base` package tries to help by listing (a subset of) relevant laws, but it's not an authoritative list of _the_ "monad laws".)

In that light, I figured that it would make sense for me to mirror the conceptual grouping of laws with two explicit records. This has the additional benefit that I can ask the question whether `IsDefaultMonad` should be merged with the `Monad` class, while keeping `MonadLaws` separate. The idea is that `IsDefaultMonad` concerns the _definitions_ of certain functions, while `MonadLaws` concerns properties that do not have the form of a definition.

(I want to note that using this conceptual grouping makes it easier to ask the question of "Is there a law for each function definition?", which in turn makes it possible to discover that `def->>->>=` is not provable from the implementation before this pull request.)

That said, I don't mind lumping `MonadLaws` and `IsDefaultMonad` together — this would reflect a different conceptual grouping along the lines of "function definitions" (`Monad`) and "any properties those functions may have" (`MonadLaws`).

omelkonian: > The identities that I had put in the IsDefaultMonad record concern the definition of functions that are traditional or relate to superclasses, but are not part of the "monad laws", such as pure, >>, fmap, and (<*>).

Then it's the wrong name :P

> That said, I don't mind lumping MonadLaws and IsDefaultMonad together 

Just to be clear, I'm not suggesting a single record with all laws, we've confused two separate orthogonal issues:
1. separating the laws about **default** methods in `IsDefaultMonad`
2. separating the monad laws into the "pure" Monad laws and interactions with superclasses

I completely agree with (1), but done in a DRY fashion.
I don't have a strong opinion for (2), but it can wait for a future PR when the separation has at least a single use.

lib/base/Haskell/Law/Monad/Def.agda
```diff
@@ -9,35 +9,70 @@ open import Haskell.Prim.Monoid
...
-    -- fmap f xs  =  xs >>= return . f
-    fmap2bind : {a b : Type} → (f : a → b) (ma : F a)
-      → fmap f ma ≡ (ma >>= (return ∘ f))
-    -- (>>) = (*>)
-    rSequence2rBind : (ma : F a) → (mb : F b) → (ma *> mb) ≡ (ma >> mb)
+    def-fmap->>= : ∀ {a b} (f : a → b) (ma : m a)
+      → fmap f ma ≡ ma >>= (return ∘ f)
+
+    def-<*>->>= : ∀ {a b} (mab : m (a → b)) (ma : m a)
+      → (mab <*> ma) ≡ (mab >>= (λ f → (ma >>= (λ x → return (f x)))))
```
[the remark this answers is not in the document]
omelkonian: Still doesn't make sense to have monad-functor laws inside the `IsDefaultMonad`.

lib/base/Haskell/Law/Monad/Def.agda
omelkonian: > trying to avoid the repetition actually decreases readability

Yeah, that can be the case quite often with any form of abstraction, but inlining code is rarely the answer.
In  your `filter` example, you've lost access to all proofs about `filter`...
I share your desire for readability, but this could be resolved at the doc-generation stage (e.g. head-reducing this definition).

> After all, the purpose of the DefaultMonad class is to avoid some form of code duplication when defining Monad class instances

This is not the only purpose of `DefaultMonad`; it is precisely the way we emulate in agda2hs (although here it doesn't apply because it's an existing class).

> (Introducing the MonadDefaults class also adds code, e.g. the type signatures for >>-from->>= and return-from-pure need to be written down — I'm not sure that the total amount of written code is reduced.)

First, `MonadDefaults` is a sensible thing to have in general, i.e. giving names to how you derive default methods from the other existing methods. Second, you've conflated two different kinds of repetition, the type repetition being a mostly harmless one, while inlining the actual body of the default program can introduce deep bugs.

jespercockx:
@HeinrichApfelmus I see you pushed some changes here, could you please give a summary of the status of this PR? Are there any remaining questions to be addressed?

omelkonian:
@jespercockx If you're fine with Monad-Functor laws being in `IsDefaultMonad` interface, feel free to merge.

HeinrichApfelmus:
I will combine `MonadLaws` and `IsDefaultMonad` into a single record `PreLawfulMonad` and then merge later if no one objects.

PR #417, opened 2025-07-22
Opinionated changes to Compile/Name for readability

liesnikov:


src/Agda2Hs/Compile/Name.hs
```diff
@@ -192,16 +196,17 @@ compileQName f
     -- _getSort is not for that; e. g. a data has the same sort as its constructor.
     getNamespace :: QName -> C (Hs.Namespace ())
     getNamespace qName = do
-      definition <- getConstInfo qName
-      case isSort $ unEl $ getResultType $ defType definition of
-        Just _         -> (reportSDoc "agda2hs.name" 25 $ text $ (prettyShow $ nameCanonical $ qnameName f)
-                                              ++ " is a type operator; will add \"type\" prefix before it") >>
-                          return (Hs.TypeNamespace ())
-        _              -> return (Hs.NoNamespace ())
+      definition <- isSort . unEl . getResultType . defType <$> getConstInfo qName
```
jespercockx, line 199: This shouldn't be named `definition` anymore but something like `qIsSort` instead

PR #424, opened 2025-09-02
Require no-eta-equality on records compiled to Haskell data

### LINKED ISSUE #320: Newtype constructor erroneously removed
Agda

```agda
record Hash (a : Set) : Set where
  constructor MakeHash
  field hashBytes : ByteString
open Hash public
{-# COMPILE AGDA2HS Hash newtype deriving (Generic, Show) #-}

instance
  iMaybeHashable : ⦃ i : Hashable a ⦄ → Hashable (Maybe a)
  iMaybeHashable {{_}} .hash Nothing = MakeHash (hashBytes (hash iUnitHashable tt))
  iMaybeHashable {{i}} .hash (Just x) = MakeHash (hashBytes (hash iPairHashable (tt , x)))
{-# COMPILE AGDA2HS iMaybeHashable #-}
```

becomes Haskell

```haskell
newtype Hash a = MakeHash{hashBytes :: ByteString}
                   deriving (Generic, Show)

instance (Hashable a) => Hashable (Maybe a) where
    hash Nothing = hash ()
    hash (Just x) = hash ((), x)
```

even though `MakeHash . hashBytes != id` since in this case we had `MakeHash :: ByteString -> Maybe a` but `hashBytes :: () -> ByteString` etc.

jespercockx:
This fixes #320 by requiring all record types that are compiled to a Haskell data type to have eta-equality disabled. It's a bit annoying but I believe it's the right thing to do since Haskell does not have a notion of eta-equality.

### DIFF
--- a/src/Agda2Hs/Compile/Record.hs
+++ b/src/Agda2Hs/Compile/Record.hs
@@ -10,7 +10,7 @@ import qualified Data.Map as Map
 
 import Agda.Compiler.Backend
 
-import Agda.Syntax.Common ( Arg(unArg), defaultArg )
+import Agda.Syntax.Common ( Arg(unArg), defaultArg, HasEta'(..) )
 import Agda.Syntax.Internal
 import Agda.Syntax.Common.Pretty ( prettyShow )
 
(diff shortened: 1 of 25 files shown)


liesnikov:
Wouldn't it make sense to enable `--no-eta-equality` by default then?

Issue #155, opened 2023-03-02
Add support for ornamented datatypes and functions


jespercockx:
When defining a fancy version of an existing datatype (e.g. `All` as a fancy version of `List`), it would be nice to be able to mark it as the ornamented version of that datatype in agda2hs, so they can both be compiled to the same Haskell datatype. The condition here would be that compiling the ornamented version should result in the exact same Haskell code as the base type (modulo naming of the type).

Similarly, we could allow marking functions (e.g. `map` on the `All` type) as an ornamented version of an existing function, and compile calls to it to calls to that existing function. Again, this would require both versions of the function to compile down to the precise same Haskell code (modulo naming of the function).

J0s3c4rl0s:
I think my comment on the Agda issue [here](https://github.com/agda/agda/issues/7701#issuecomment-2674917209) could be useful in this context. 

I also think this would obviate the need for agda2hs' "boxed" pragma right? As you could just define types to compile to some simpler type and thus not need to stick to ornament-esque "boxed" types with data + proof in a sigma type.

jespercockx:
You mean the `unboxed` pragma? I'm not sure if it would be subsumed by this feature: basically, this feature would allow you to drop constructor *arguments* and datatype *indices*, while `unboxed` allows you to drop the *constructor* itself (if there is a single constructor with a single non-erased argument). But perhaps it's possible to design a feature that encompasses both.

anuyts:
Is it correct that the underlying goal would be to:
- reduce the proliferation of Haskell definitions,
- optimize (e.g. compile certain functions to the identity)?

jespercockx:
I think that's one way to frame it, yes. The other way would be to see it as making it much *easier* to define Agda datatypes and functions with different invariants and pre/postconditions without creating new Haskell-side code. The hope would be that once you have this ability you would be more likely to introduce these new types in the first place (as opposed to taking an extrinsic verification approach).

PR #430, opened 2025-10-03
[ #155 ] Add `COMPILE ... to` pragmas for data types and functions

jespercockx:
This is a first attempt at implementing a `COMPILE ... to` pragma for compiling a datatype or function to an existing Haskell definition, as described in #155. Currently there are the following limitations:

* It makes use of an `IORef` to keep track of the mapping between names, which hopefully should not be a problem.
* When checking a `COMPILE ... to` pragma, despite generating no code it still compiles both the definition in the pragma *and* the target definition, which is a bit wasteful. It would perhaps be more efficient to store the compiled definitions in another `IORef` so we don't need to recompile them later (there are other places in the agda2hs codebase where this would be useful too).
* Type classes, type aliases, and anything that's not a regular datatype or a function with a normal `COMPILE` pragma is not yet supported.
* I only did minimal testing so far.

I'll probably add some more tests (including failing ones) later and perhaps add support for type aliases too, but I'm already curious to see what people think about this.


J0s3c4rl0s:
Yes this sounds interesting, I could use some of the agda examples in `src/AgdaRunIdExamples.agda` from my runtime_id_fn repo I used for my thesis to test the feature.

PR #439, opened 2025-11-03
[ fix #306 ] Correctly error on dot patterns corresponding to module arguments

### LINKED ISSUE #306: Dot patterns in arguments corresponding to module parameters should not be allowed
```agda
open import Haskell.Prelude

module _ where

cast : {a b : Set} → @0 a ≡ b → a → b 
cast {a} {b} = cast'
  where
  cast' : @0 a ≡ b → a → b
  cast' refl x = x

{-# COMPILE AGDA2HS cast #-}
```
This is currently accepted, but compiles to invalid Haskell code:
```haskell
cast :: a -> b
cast = cast'
  where
    cast' :: a -> b
    cast' x = x
```

NathanielB123:
We delay applying `pars` from `getContextArgs` until after we have checked for dot patterns.
Relaxes the dot pattern error to only apply to arguments compiled to (implicit or explicit) `forall`s

Also stop throwing `__IMPOSSIBLE__` on instances inside modules.

### DIFF
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -37,7 +37,7 @@ import Agda.Utils.Size ( Sized(size) )
 
 import Agda2Hs.AgdaUtils
 import Agda2Hs.Compile.Name ( compileQName )
-import Agda2Hs.Compile.Term ( compileTerm, usableDom )
+import Agda2Hs.Compile.Term ( compileTerm, usableDom, dependentDom )
 import Agda2Hs.Compile.Type ( compileType, compileDom, DomOutput(..), compileDomType )
 import Agda2Hs.Compile.TypeDefinition ( compileTypeDef )
 import Agda2Hs.Compile.Types
@@ -194,7 +194,7 @@ compileClause' curModule pars projName x ty c = do
   -- Check whether any parameters (including module parameters) are forced
   -- See https://github.com/agda/agda2hs/issues/306
   annPats <- annPats ty (namedClausePats c)
-  traverse (uncurry checkNonErasedForced) annPats
+  traverse (uncurry checkIllegalForced) annPats
   ty' <- piApplyM ty pars
   compileClause'' curModule projName x ty' (c `apply` pars)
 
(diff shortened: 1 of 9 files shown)


jespercockx:
Thank you very much for the PR!

NathanielB123:
Oops, should be fixed now

src/Agda2Hs/Compile/Function.hs
```diff
@@ -271,15 +297,14 @@ compilePats ty ((namedArg -> pat):ps) = do
   (a, b) <- mustBePi ty
   reportSDoc "agda2hs.compile.pattern" 10 $ text "Compiling pattern:" <+> prettyTCM pat
   let rest = compilePats (absApp b (patternToTerm pat)) ps
-  when (usableDom a) checkForced
+  checkForced a pat
```
jespercockx: Why did you remove the condition `usableDom a` here? Dot patterns in erased positions should still be allowed (including for erased parameters).

NathanielB123: I moved the `when (usableDom a)` into `checkForced` (note that `checkForced` now takes the type `a` as well as `pat`)

NathanielB123: The problem is that the pattern-matching lambda in `checkSubst` forces a relevant dot pattern on `t`
```agda
checkSubst : ∀ {@0 x α β} (t : Telescope α (x ∷ β)) → Bool
checkSubst t = caseTelBind t λ ty rest → λ where refl → True
```
i.e. if I manually translate the lambda into a top-level definition then we have
```agda
checkSubst' : ∀ {@0 x α β} (t : Telescope α (x ∷ β)) (ty : Bool) (rest : Telescope (x ∷ α) β) → @0 t ≡ ExtendTel ty rest → Bool
checkSubst' .(ExtendTel ty rest) ty rest refl = True

checkSubst : ∀ {@0 x α β} (t : Telescope α (x ∷ β)) → Bool
checkSubst t = caseTelBind t (checkSubst' t)
```
I guess the point which  could be made here is that sometimes the dot pattern isn't actually required for the translated Haskell code to typecheck, but I'm not what a good way to detect this would look like...

jespercockx: Perhaps a criterion that we could use is that there should not be a dot pattern for arguments that are compiled to an (implicit or explicit) `forall` in the Haskell code, but for regular term-level arguments it should be fine since these cannot be dependent in Haskell.

jespercockx: Thank you, I'm happy with that solution!

PR #440, opened 2025-11-04
Compile dot patterns

NathanielB123:
As of #439 it is actually correct behaviour for some clauses with dot patterns to get compiled to Haskell code (specifically when the argument is compiled to a term argument in Haskell - i.e. there is no dependent quantification).

Names given by the Agda programmer are retained where possible, but explicit dot patterns are just turned into wildcards

OLD:
Arguably it might be nice to retain names (e.g. prefixed by a `_`) when the dot pattern isn't provided explicitly, but this info does not appear to be retained by the time we reach internal syntax (I could certainly be missing something though)

### DIFF
--- a/src/Agda2Hs/Compile/Function.hs
+++ b/src/Agda2Hs/Compile/Function.hs
@@ -37,7 +37,7 @@ import Agda.Utils.Size ( Sized(size) )
 
 import Agda2Hs.AgdaUtils
 import Agda2Hs.Compile.Name ( compileQName )
-import Agda2Hs.Compile.Term ( compileTerm, usableDom, dependentDom, extractMaybeName )
+import Agda2Hs.Compile.Term ( compileTerm, usableDom, dependentDom )
 import Agda2Hs.Compile.Type ( compileType, compileDom, DomOutput(..), compileDomType )
 import Agda2Hs.Compile.TypeDefinition ( compileTypeDef )
 import Agda2Hs.Compile.Types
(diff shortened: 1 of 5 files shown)


NathanielB123:
Yes good point - I did not look very carefully, my bad. The naming behaviour on this PR is now:

- Dot pattern with no explicit name: becomes a wildcard `_`
- Dot pattern with explicit name/inferred dot pattern: `_` prefixed to the given name
- Dot pattern from module param: module param name used

I like the idea of prefixing with `_` if possible, so I think ideally dotted module parameters would also be named like this, but it is a bit awkward because the code that compiles module parameters currently does this entirely based off the type. Also there is the potential issue that the Agda programmer could prefix a parameter name with an `_`, so modifying the names could then cause clashes, e.g.
```
checkSubst' : ∀ {@0 x α β} (t : Telescope α (x ∷ β)) (ty : Bool) (rest : Telescope (x ∷ α) β) → @0 t ≡ ExtendTel ty rest → Bool
checkSubst' t _t rest refl = True
```
compiles to invalid Haskell
```
checkSubst' :: Telescope -> Bool -> Telescope -> Bool
checkSubst' _t _t rest = True
```
Because of this, I'm somewhat inclined to just use the names as-given (easier, and lower chances of weird edge cases)

jespercockx:
I thought perhaps with your changes we would get rid of `compileModuleParams` completely and just compile them as regular function arguments for each clause, but if we keep that function then indeed what you're saying makes sense.

PR #455, opened 2026-03-06
Support for equality constraints

zeroeightysix:
This PR adds rudimentary (but unfortunately, not very useful - at the moment) support for generating equality constraints.

Any occurrence of a non-erased propositional equality in an instance argument is treated as being an equality constraint. The following compiles as such:
```agda
f : ⦃ a ≡ b ⦄ → c
```
```haskell
f :: a ~ b => c
```

The reason I say this isn't very useful yet is because you're not allowed to actually learn anything from the equality yet, due to #168.

(If the changes are okay, best to squash, as the git history isn't the cleanest)


src/Agda2Hs/Compile/Type.hs
```diff
@@ -250,6 +253,16 @@ compileDomType x a =
         | otherwise -> return $ DomForall Nothing
     DOTerm     -> fmap (uncurry DomType) . withNestedType . compileTypeWithStrictness . unEl $ unDom a
 
+compileEqualityConstraint :: Term -> C CompiledDom
+compileEqualityConstraint t = do
+  Def _ es <- reduce t
+  -- The arguments to equality are _a_ (level), _A_ (the type of elements), x (: A), and y (: A)
+  -- We want to compile x and y
+  let Just (_:_:x:y:_) = allApplyElims es
+  hsX <- compileType (unArg x)
```
jespercockx, line 262: Since we are compiling `x` and `y` as types, should we have a check that their type `A` is indeed compiling to a Haskell kind?

jespercockx:
Thank you for the PR! I'm happy with the feature but I'm not yet convinced that it is really necessary to treat equality constraints different from other (class) constraints in the translation. In particular, I don't really see why we need new constructors `DomEquality` and `DOEquality` instead of just using the ones for constraints, could you explain the reasoning behind that?

Issue #126, opened 2022-11-14
Handle public imports properly


jespercockx:
Currently, agda2hs handles transitive imports by importing the names directly from the module where they were originally defined, ignoring any re-exporting layers in the Agda code. We should instead preserve the import structure that the user picked. Note that this would first require us to handle export lists (see https://github.com/agda/agda2hs/issues/44), which are Haskell's way of doing public imports.

PR #456, opened 2026-03-18
Refactor base library structure and port additional Haskell functions

loki259:
## Summary

This PR reorganizes large parts of the Agda2Hs base library to better separate primitive definitions from higher-level utilities and to align the module structure more closely with the hierarchy found in the Haskell library.

In addition to the structural refactoring, several modules were expanded with additional definitions ported from Haskell.

## Main changes

- Introduced new modules mirroring the Haskell version of the library:

  - Added `Data.String`
  - Added `Data.Functor`
  - Added `Data.Foldable`
  - Added `Data.Traversable`
  - Added `Control.Applicative`

- Refactored `Prim` modules so they mainly contain class definitions and minimal primitive functionality:

  - Moved `_<$>_`, `_<&>_`, `_$>_`, `void` from `Prim.Functor` to `Data.Functor`
  - Moved `_=<<_`, `mapM₋`, `sequence₋` from `Prim.Monad` to `Control.Monad` and `Data.Foldable` and moved `bindIO` from `Prim.Monad` to `Prim.IO`
  - Moved `lines`, `words`, `unlines`, `unwords` from `Prim.String` to `Data.String`
  - Moved all functions in `Prim.List` except `_++_`, `head`, `last`, `tail`, `init`, `map` to `Data.List`
  - Moved `lengthNat` from `Prim` to `Prim.List`
  - Moved `reverse`, `_!!ᴺ_`, `_!!_`, `lookup` from `Prelude` to `Data.List`

- Added additional type classes and instances and refactored existing type class
  definitions:

  - Reworked `IO` instances (`Functor`, `Applicative`, `Monad`) so they are now defined explicitly in terms of `bindIO` and `returnIO`.
  - Added `Alternative` class record and instances (defined in new module `Prim.Alternative`)
  - Added `MonadPlus` class record and instances (defined in new module `Prim.MonadPlus`)
  - Moved `MonadFail` from `Prim.Monad` into its own module `Prim.MonadFail` and added instance for `IO`
  - Added `MonoidFirst` and `MonoidLast` `Monoid` instances in `Prim.Monoid`
  - Refactored `Applicative` to include `liftA2` as additional method (with corresponding default record)
  - Refactored the method set of `Foldable` to match the Haskell version (moving `any`, `all`, `and`, `or`, `concat`, `concatMap`, `notElem` to `Data.Foldable` and adding `fold`, `foldMap'`, `foldr'`, `foldl'`, `foldr1`, `foldl1`, `maximum`, `minimum`)
  - Moved `Semigroup` from `Prim.Monoid` into its own module `Prim.Semigroup`
  - Introduced submodules for instances of type classes

- Expanded several modules with additional definitions ported from Haskell (still missing definitions are listed in a comment at the end of the respective file):

  - Expanded `Control.Monad` (also changed signature of existing `guard` definition to match the Haskell version)
  - Expanded `Data.List` (also updated the definition of `nub` and removed some property proofs based on the outdated definition)
  - Ported `Data.Functor`
  - Ported `Data.Foldable`
  - Ported `Data.Traversable`
  - Ported `Control.Applicative`
  - Expanded `Prelude` by updating module imports (also updated documentation of missing definitions from the Haskell prelude)

- Updated imports and reexports throughout the codebase to reflect the new module structure

## Motivation

The goal of these changes is to make the structure of the library more consistent with the organization familiar from Haskell, while keeping the `Prim` modules focused on primitive definitions. This separation also makes it easier to extend the library with additional standard definitions in the future and avoids many cyclic module dependencies when reusing definitions from other parts of the library.

## Notes

Because many of the refactorings depend on each other (often due to running into problems with cyclic imports), the changes are grouped into a single commit.


lib/base/Haskell/Control/Applicative.agda
```diff
@@ -0,0 +1,29 @@
...
+liftA3 f a b c = liftA2 f a b <*> c
+
+optional : ⦃ Alternative f ⦄ → f a → f (Maybe a)
+optional v = Just <$> v <|> pure Nothing
+
+-- Omitted for now:
+-- - 'newtype Const a (b :: k) = Const { getConst :: a }'
+-- - 'newtype WrappedMonad (m :: Type -> Type) a = WrapMonad { unwrapMonad :: m a }'
+-- - 'newtype WrappedArrow (a :: Type -> Type -> Type) b c = WrapArrow { unwrapArrow :: a b c }'
+-- - 'newtype ZipList a = ZipList { getZipList :: [a] }'
```
jespercockx: Could you please ensure that each files conforms to the standard git whitespace conventions, in particular have an empty newline at the end of the file? (I use the "Trailing Whitespace" plugin for VS Code to fix these issues automatically).

jespercockx: We could also add an .editorconfig file like the one that was added to the main Agda repo, if you think that would help (https://github.com/agda/agda/blob/master/.editorconfig).

lib/base/Haskell/Control/Monad.agda
```diff
@@ -2,10 +2,138 @@ module Haskell.Control.Monad where
...
+open import Haskell.Prim.Monad public
+open import Haskell.Prim.MonadFail public
+open import Haskell.Prim.MonadPlus public
+open import Haskell.Prim.Traversable public using (mapM; sequence)
+open import Haskell.Data.Traversable public using (forM)
+open import Haskell.Data.Foldable public using (mapM₋; forM₋; sequence₋; msum)
+open import Haskell.Data.Functor public using (void)
+
+
+variable a1 a2 a3 a4 a5 r : Type
```
jespercockx: Could you please mark these variables as `private` to avoid polluting the namespace too much?

lib/base/Haskell/Control/Monad.agda
```diff
@@ -2,10 +2,138 @@ module Haskell.Control.Monad where
...
+
+join : ⦃ Monad m ⦄ → m (m a) → m a
+join x = x >>= id
+
+mfilter : ⦃ MonadPlus m ⦄ → (a → Bool) → m a → m a
+mfilter p ma = do
+  a ← ma
+  if p a then return a else mzero
+
+filterM : ⦃ Applicative m ⦄ → (a → m Bool) → (List a) → m (List a)
```
jespercockx: No need for parentheses around `List a` (same for the functions below).

lib/base/Haskell/Control/Monad.agda
```diff
@@ -2,10 +2,138 @@ module Haskell.Control.Monad where
...
+filterM p = foldr (λ x → liftA2 (λ b → if b then (x ∷_) else id) (p x)) (pure [])
+
+mapAndUnzipM : ⦃ Applicative m ⦄ → (a → m (b × c)) → (List a) → m (List b × List c)
+mapAndUnzipM f xs = fmap unzip (traverse f xs)
+
+zipWithM : ⦃ Applicative m ⦄ → (a → b → m c) → (List a) → (List b) → m (List c)
+zipWithM f xs ys = sequenceA (zipWith f xs ys)
+
+zipWithM₋ : ⦃ Applicative m ⦄ → (a → b → m c) → (List a) → (List b) → m ⊤
```
jespercockx: What is the symbol `₋` you're using here? Have you made sure it compiles to the regular `_` in Haskell?

lib/base/Haskell/Data/List.agda
```diff
@@ -1,90 +1,310 @@
...
+
+-- Omitted for now:
+-- [obviously non-terminating]
+-- - 'iterate :: (a -> a) -> a -> [a]'
+-- - 'iterate' :: (a -> a) -> a -> [a]'
+-- - 'repeat :: a -> [a]'
+-- - 'cycle :: [a] -> [a]'
+-- - 'unfoldr :: (b -> Maybe (a, b)) -> b -> [a]
+
+-- [hard to prove termination]
```
jespercockx, line 294: Maybe instead of implementing these in Agda, we could instead expose them as `postulates`, perhaps together with some postulates of their properties (but that's not needed for this PR).

lib/base/Haskell/Data/String.agda
```diff
@@ -0,0 +1,40 @@
...
+
+open import Haskell.Prim
+open import Haskell.Prim.List
+open import Haskell.Prim.Foldable
+open import Haskell.Data.Foldable
+
+open import Haskell.Prim.String public
+
+
+private
```
jespercockx: Instead of marking these as `private`, could we put them inside a submodule? That way they don't pollute the namespace but users can still refer to these helper functions for the purpose of writing proofs.

lib/base/Haskell/Data/Traversable.agda
```diff
@@ -0,0 +1,101 @@
...
+open import Haskell.Prim hiding (s)
+open import Haskell.Prim.Functor
+open import Haskell.Prim.Applicative
+open import Haskell.Prim.Monad
+open import Haskell.Prim.Tuple
+
+open import Haskell.Prim.Traversable public
+
+
+variable s : Type
```
jespercockx: Please mark this one private too.

lib/base/Haskell/Data/Traversable.agda
```diff
@@ -0,0 +1,101 @@
...
+
+for : ⦃ Traversable t ⦄ → ⦃ Applicative f ⦄ → t a → (a → f b) → f (t b)
+for = flip traverse
+
+forM :  ⦃ Traversable t ⦄ → ⦃ Monad m ⦄ → t a → (a → m b) → m (t b)
+forM = flip mapM
+
+private
```
jespercockx: Why are these marked as `private`? I think they should just be public, no?

jespercockx (changes requested):
Thank you for the nice PR! Generally things look good but I have a few small requests for changes. Also, could you please investigate the failure on CI (at least the non-Nix ones).

lib/base/Haskell/Control/Applicative.agda
loki259: Sorry, I have fixed the whitespace issue by now. Still working on fixing the tests (which proved harder than expected as I described on Zulip).
I am actually a bit embarrassed that I only used the Agda compiler and didn't think of running the checks that are implemented in the makefile but here we are.
The makefile also helped me to fix the whitespace issue btw so I don't think an .editorconfig is really needed.

lib/base/Haskell/Control/Monad.agda
loki259: Before my refactoring there were already at least two functions (`mapM₋` and `sequence₋`) in Prim.Monad with this symbol in place of Haskell's underscore and I just copied that naming scheme for the other result-discarding functions from Haskell that I ported.
I must admit that I didn't think about how it is compiled though to be honest. How can I make sure that the symbol gets properly converted?

lib/base/Haskell/Data/List.agda
loki259: That's a good idea 👍

lib/base/Haskell/Data/Traversable.agda
loki259: They are not part of the Haskell module and I also didn't know where else to put their definitions since `State` and `StateT` are defined as part of the transformers package to my knowledge. They are essentially only used as helpers here, so that's why I decided to make them private in the end. Do you think there is a better approach (other than porting the transformers package as well)?

lib/base/Haskell/Data/String.agda
loki259: I can do that. How would you name that submodule?

lib/base/Haskell/Control/Monad.agda
jespercockx: Actually it seems that agda2hs automatically converts `₋` into `_`, see implementation of `hsName`. So what you did is correct, no changes needed.

lib/base/Haskell/Data/String.agda
jespercockx: Something like `StringHelpers` perhaps?

lib/base/Haskell/Data/Traversable.agda
jespercockx: Well the Haskell implementation of these functions does not use the transformers package, or does it? I think we should just try to mirror that implementation.

lib/base/Haskell/Data/String.agda
loki259: Sounds good!
But just to be clear: Do you want me to add StringHelpers as a sibling module next to Data.String in the hierarchy or do you want a submodule inside Data.String?
In the latter case I would suggest calling it just "Helpers" since it is already part of the "String" module.

lib/base/Haskell/Data/Traversable.agda
loki259: I looked at the GHC implementation and they are using similar State monad definitions from an internal module (GHC.Internal.Data.Functor.Utils). That is why I also wanted to make it "internal" by having the definitions be private. We can of course also move them to some internal module. On the other hand they are only used in Data.Traversable at the moment, so maybe that refactoring can wait? What do you say?

loki259:
> Also, could you please investigate the failure on CI (at least the non-Nix ones).

I could use some help in fixing the issues responsible for the remaining test failures (as already posted on Zulip):

When I reexport definitions in another module (which I did with a lot of the newly ported functions that are part of the Prelude but needed to be defined elsewhere) the Haskell files produced by Agda2Hs seem to import those definitions from the module where they were originally exported instead of from the module where the corresponding Agda file imports them from.
Specifically, there are some tests that fail now because while the Agda file imports everything from the Prelude the compiled Haskell version of the file imports identifiers from Control.Monad or Data.List for example (although the relevant identifiers are reexported from the Prelude and importable from there on the Agda side).
Is there something I am not aware of when it comes to how reexports are handled in Agda2Hs (or in Agda generally)? I read the imports section in the documentation but I couldn't find anything on this matter.

Another issue I have is that type class instances are seemingly not reexported unless I list the instances explicitly along with the class definition in the using clause of the public import. Listing them is of course not the end of the world but it will make adding new instances a bit cumbersome in the future because the import lists for the respective module need to be adjusted wherever the new instance should also be reexported.
Is there a way around this? Can I somehow explicitly export the instances all at once?

lib/base/Haskell/Data/String.agda
jespercockx: Yes you are right. Both solutions would be fine with me, but in this case a separate file feels a bit overkill to me.

lib/base/Haskell/Data/Traversable.agda
jespercockx: I'm fine with leaving the refactoring to a future time when we also need the code in other places, either having `private` definitions or a nested `Helpers` module would be fine for now.

jespercockx:
You're not missing anything, proper re-exports are still on the todo-list (see https://github.com/agda/agda2hs/issues/126). As a workaround I guess you could replace the re-exports by postulates for now, possibly together with an equality proof that these postulates are equal to the real versions (this is not a very nice solution but I can't really think of a better one atm).

For exporting instances, I don't think it should be necessary to explicitly import them at all, just importing the module should be sufficient to bring them in scope for instance resolution on the Agda side. Or is the problem that they are somehow not exported on the Haskell side?

lib/base/Haskell/Data/Traversable.agda
loki259: I introduced a Helpers submodule to be in line with the approach for Data.String 👍

loki259:
Oh okay, I didn't see that there was already an issue for this reexport problem.
Your postulates workaround makes sense (although I am not sure where to put those equality proofs) but wouldn't it be easier to just import the relevant definitions qualified and to then define properly exported aliases to them? For example, instead of having `open import Haskell.Data.List public using (or)` in the Prelude module we could add
```agda
or : ⦃ Foldable t ⦄ → t Bool → Bool
or = Haskell.Data.List.or
```
This avoids postulates and seems a bit simpler to me. What do you think?

Regarding the instance exports: There is no problem on the Haskell side but on the Agda side I cannot just use `open import Haskell.Control.Monad` to make use of the module's `guard` function, for example, because it depends on instances of the `Alternative` class being in scope. Those instances are imported in Control.Monad but apparently they are not reexported by default.
I can of course add a second import statement that exposes the instances from Prim.Alternative directly to the Agda file, but I think it would be cleaner if the Agda version would only need to import modules of functions that it uses (not modules where instances are defined, just like the Haskell version).
I already validated that I can achieve that by explicitly reexporting all instances of the `Alternative` class from Control.Monad but I was wondering if there is a better way to go about it since this approach forces us to update the list of exported instances in multiple files whenever a new instance is added in the source module. Can I somehow export all the instances at once?

jespercockx:
> This avoids postulates and seems a bit simpler to me. What do you think?

Yes, if that works it's a much better solution!

> I already validated that I can achieve that by explicitly reexporting all instances of the Alternative class from Control.Monad but I was wondering if there is a better way to go about it since this approach forces us to update the list of exported instances in multiple files whenever a new instance is added in the source module. Can I somehow export all the instances at once?

I believe it should also be sufficient to just give a name to the module that you want to re-export the instances from, e.g.

```agda
import Prim.Alternative
module Alternative = Prim.Alternative
```

Could you give that approach a try to see if if works?

loki259:
Unfortunately, my approach of introducing explicit alias definitions in place of reexports didn't work as well as expected: 
There seem to be test cases where Agda2Hs introduces a qualified Prelude import for some definitions although the Prelude is already imported implicitly. This happens even when there is already another qualified Prelude import.
There is also a bigger problem: In contrast to using Agda's builtin reexport mechanism, the explicit aliasing approach creates import conflicts in terms of name collisions on the Agda side whenever a different module with a subset of the definitions from the Prelude is also imported (for example when importing the Prelude and Control.Monad).
From what I can see these two problems are responsible for about half of the current test failures. There are, however, also a few tests where I have really no idea what broke them and why certain definitions that are clearly exported from the Prelude seem to no longer be in scope in these files.
Anyways, even if I was able to identify what causes these other tests to fail, I don't think we can make the explicit reexports work without fixing some of the Agda2Hs internals related to #126 first. This conclusion probably also holds for your suggestion of using postulates. Maybe you can take a look at the failing test cases yourself and tell me if you agree with my assessment.

On the positive side, I found a solution for the instance export problem:
I first implemented your suggestion of using named modules when importing type class modules but that created a lot of name conflicts and brought me to the idea of instead exporting the instances as part of Instances submodules on the side of the type class modules. This avoids the need for finding a new name for the whole type class module whenever definitions need to be imported and one can instead just explicitly import the Instances submodule (optionally renaming that submodule in case of conflicts with Instances submodules from other type classes).

jespercockx:
Ah, that is unfortunate. It seems this is blocked on implementing re-exports properly then. I would like to look at it but I also have a few open issues in Agda itself I need to fix first for the release of 2.9, so I'm not sure when I'll be able to get to it. In case anyone else want to give it a go, I would be happy to provide some pointers.

